PR #4, opened 2015-02-24
Replace direct socket junk with a dispatcher

jlee-r7:
### Verification
 - [ ] Have a Windows Vista+ box (below example is at 192.168.100.156)
 - [ ] Works as before:
```ruby
d = Smb2::Dispatcher::Socket.connect("192.168.100.156", 445)
c = Smb2::Client.new(dispatcher: d, username:"administrator", password:"P@ssword1")
c.negotiate
c.authenticate
# => 0
```


trosen-r7:
What else is going into the "Dispatcher" namespace?

Issue #14, opened 2015-04-15
Client: Create file for reading


jlee-r7:
something like this:
```ruby
tree_id = client.tree_connect("\\\\hostname\\share")
client.read(tree_id, "file.txt")
# => "contents of file.txt"
```
needs to 
- send a `CreateRequest` with appropriate values
- read the response, store the `file_id` therein
- send a `GetInfo` request for that `file_id`
- read the response, store the size
- send a `ReadRequest` for that many bytes
- return the value as a String

limhoff-r7:
Is the `tree_id` valid on other clients or is it client specific?  If it's client specific, wouldn't it make more sense for `client.tree_connect` to return a `Tree` object and have it have `Tree.read` as the client is implied by the returned `tree`.

jlee-r7:
Oh, that's a good idea.  Yes, tree IDs are client-specific and Create will be tree-specific.

jlee-r7:
Current thinking looks like this:
```ruby
tree = client.tree_connect("\\\\hostname\\share") # => #<Smb2::Tree ... >
file = tree.create("\\path\\to\\file.txt") # => #<Smb2::File ... >
file.read # => "contents of file.txt"
```

jlee-r7:
fixed in #28

PR #28, opened 2015-04-16
Read files

jlee-r7:
This works with code like this:

# Verification Steps
- [x] `bundle install`
- [x] `bundle console`
- [x] `require 'smb2'`
- [x]  Get a `host` name from either a local VM or one on the network with SMB2 (so Vista or newer).
- [x] `dispatcher = Smb2::Dispatcher::Socket.connect(host, 445)`
- [x] Get the username and password for a local administrator on the host
- [x] `client = Smb2::Client.new(dispatcher: dispatcher, username: username, password: password)`
- [x] `client.negotiate`
- [x] `client.authenticate`
- [x] Log onto the machine using redesktop and create a file, `C:\Users\Public\foo.txt`, with some content for reading.
- [x] `tree = client.tree_connect("\\\\#{host}\\C$")`
- [x] `file = tree.create("users\\public\\foo.txt".encode('utf-16le'))`
- [x] VERIFY `puts file.read offset: 0, length: 10` is first 10 bytes of the expected content.
- [x] VERIFY `file.read.length` is the size of the file
- [x] VERIFY `Digest::SHA1.hexdigest(data)` matches the sha1sum of the file using theses instrutions to calculate the SHA1 using Powershell: http://learn-powershell.net/2013/03/25/use-powershell-to-calculate-the-hash-of-a-file/.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - '2.0'
   - '2.1'
   - '2.2'
-  - 'jruby-19mode'
+  - 'jruby-9000'
--- a/README.md
+++ b/README.md
@@ -30,17 +30,7 @@ Or install it yourself as:
 
 ### Using the `Client` class
 
-```ruby
-sock = TCPSocket.new("192.168.100.140", 445)
-c = Smb2::Client.new(
-  socket: sock,
-  username:"administrator",
-  password:"P@ssword1",
-  domain:"asdfasdf"
-)
-c.negotiate
-c.authenticate
-```
+{include:Smb2::Client}
 
 ### Making packets manually
 
@@ -60,9 +50,14 @@ neg_response = Smb2::Packet::NegotiateResponse.new(data)
 
 ```
 
+## License
+
+{include:file:LICENSE.txt}
+
+
 ## Contributing
 
-1. Fork it ( https://github.com/jlee-r7/smb2/fork )
+1. Fork it ( https://github.com/rapid7/smb2/fork )
 2. Create your feature branch (`git checkout -b my-new-feature`)
 3. Commit your changes (`git commit -am 'Add some feature'`)
 4. Push to the branch (`git push origin my-new-feature`)
--- a/README.md
+++ b/README.md
@@ -30,7 +30,17 @@ Or install it yourself as:
 
 ### Using the `Client` class
 
-{include:Smb2::Client}
+```ruby
+sock = TCPSocket.new("192.168.100.140", 445)
+c = Smb2::Client.new(
+  socket: sock,
+  username:"administrator",
+  password:"P@ssword1",
+  domain:"asdfasdf"
+)
+c.negotiate
+c.authenticate
+```
 
 ### Making packets manually
 
@@ -52,7 +62,7 @@ neg_response = Smb2::Packet::NegotiateResponse.new(data)
 
 ## License
 
-{include:file:LICENSE.txt}
+Smb2 is released under a 3-clause BSD license. See [LICENSE.txt](LICENSE.txt) for full text.
 
 
 ## Contributing
--- a/examples/read_file.rb
+++ b/examples/read_file.rb
@@ -16,11 +16,14 @@ c = Smb2::Client.new(dispatcher: d, username:"msfadmin", password:"msfadmin")
 
 c.negotiate
 c.authenticate
-tree = c.tree_connect("\\\\#{host}\\C$")
+tree = c.tree_connect("\\\\#{host}\\Users")
 
-file = tree.create("\\autoexec.bat")
+file = tree.create("public\\foo.txt".encode('utf-16le'))
 
 # Right now this returns a Smb2::Packet::ReadResponse. When it works
 # correctly, it will be the actual file contents
-p file.read
+response = file.read
 
+puts response.header.nt_status.to_s(16)
+
+p response
--- a/examples/read_file.rb
+++ b/examples/read_file.rb
@@ -20,10 +20,4 @@ tree = c.tree_connect("\\\\#{host}\\Users")
 
 file = tree.create("public\\foo.txt".encode('utf-16le'))
 
-# Right now this returns a Smb2::Packet::ReadResponse. When it works
-# correctly, it will be the actual file contents
-response = file.read
-
-puts response.header.nt_status.to_s(16)
-
-p response
+puts file.read
--- a/examples/read_file.rb
+++ b/examples/read_file.rb
@@ -1,6 +1,6 @@
 #!/usr/bin/env ruby
 
-$:.push File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib'))
+require 'bundler/setup'
 
 require 'smb2'
 
@@ -18,6 +18,6 @@ c.negotiate
 c.authenticate
 tree = c.tree_connect("\\\\#{host}\\Users")
 
-file = tree.create("public\\foo.txt".encode('utf-16le'))
+file = tree.create("public\\foo.txt")
 
 puts file.read
--- a/examples/read_file.rb
+++ b/examples/read_file.rb
@@ -1,0 +1,26 @@
+#!/usr/bin/env ruby
+
+$:.push File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib'))
+
+require 'smb2'
+
+host = ARGV.first
+if host.nil?
+  $stderr.puts("Usage: #{$0} <ip address>")
+  exit 1
+end
+
+d = Smb2::Dispatcher::Socket.connect(host, 445)
+
+c = Smb2::Client.new(dispatcher: d, username:"msfadmin", password:"msfadmin")
+
+c.negotiate
+c.authenticate
+tree = c.tree_connect("\\\\#{host}\\C$")
+
+file = tree.create("\\autoexec.bat")
+
+# Right now this returns a Smb2::Packet::ReadResponse. When it works
+# correctly, it will be the actual file contents
+p file.read
+
--- a/lib/smb2.rb
+++ b/lib/smb2.rb
@@ -2,10 +2,12 @@
 #
 # [[MS-SMB2] Server Mesage Block (SMB) Protocol Versions 2 and 3](https://msdn.microsoft.com/en-us/library/cc246482.aspx)
 module Smb2
-  autoload :VERSION, 'smb2/version'
-  autoload :Packet, 'smb2/packet'
-  autoload :Dispatcher, 'smb2/dispatcher'
   autoload :Client, 'smb2/client'
+  autoload :Dispatcher, 'smb2/dispatcher'
+  autoload :Packet, 'smb2/packet'
+  autoload :Tree, 'smb2/tree'
+  autoload :File, 'smb2/file'
+  autoload :VERSION, 'smb2/version'
 
   # [[MS-SMB2] 2.2 Message Syntax](https://msdn.microsoft.com/en-us/library/cc246497.aspx)
   COMMANDS = {
--- a/lib/smb2.rb
+++ b/lib/smb2.rb
@@ -4,9 +4,9 @@
 module Smb2
   autoload :Client, 'smb2/client'
   autoload :Dispatcher, 'smb2/dispatcher'
+  autoload :File, 'smb2/file'
   autoload :Packet, 'smb2/packet'
   autoload :Tree, 'smb2/tree'
-  autoload :File, 'smb2/file'
   autoload :VERSION, 'smb2/version'
 
   # [[MS-SMB2] 2.2 Message Syntax](https://msdn.microsoft.com/en-us/library/cc246497.aspx)
--- a/lib/smb2/client.rb
+++ b/lib/smb2/client.rb
@@ -167,18 +167,6 @@ class Smb2::Client
     # XXX do we need the Server GUID?
   end
 
-  # @param tree [String]
-  def tree_connect(tree)
-    packet = Smb2::Packet::TreeConnectRequest.new do |request|
-      request.tree = tree.encode("utf-16le")
-    end
-
-    response = send_recv(packet)
-    response_packet = Smb2::Packet::TreeConnectResponse.new(response)
-
-    Smb2::Tree.new(client: self, tree_connect_response: response_packet)
-  end
-
   # Adjust `request`'s header with an appropriate sequence number and session
   # id, then send it and wait for a response.
   #
@@ -198,6 +186,19 @@ class Smb2::Client
     dispatcher.send_packet(request)
 
     dispatcher.recv_packet
+  end
+
+  # @param tree [String]
+  # @return [Smb2::Tree]
+  def tree_connect(tree)
+    packet = Smb2::Packet::TreeConnectRequest.new do |request|
+      request.tree = tree.encode("utf-16le")
+    end
+
+    response = send_recv(packet)
+    response_packet = Smb2::Packet::TreeConnectResponse.new(response)
+
+    Smb2::Tree.new(client: self, tree_connect_response: response_packet)
   end
 
   protected
--- a/lib/smb2/client.rb
+++ b/lib/smb2/client.rb
@@ -1,4 +1,5 @@
 require 'smb2/packet'
+require 'smb2/tree'
 require 'net/ntlm'
 require 'net/ntlm/client'
 
@@ -103,9 +104,6 @@ class Smb2::Client
         @security_mode & DEFAULT_SECURITY_MODE
       ),
     )
-    header = packet.header
-    header.command_seq = @sequence_number += 1
-    packet.header = header
 
     @ntlm_client = Net::NTLM::Client.new(
       username,
@@ -118,9 +116,7 @@ class Smb2::Client
 
     packet.security_blob = gss_type1(type1.serialize)
 
-    dispatcher.send_packet(packet)
-
-    response = dispatcher.recv_packet
+    response = send_recv(packet)
     response_packet = Smb2::Packet::SessionSetupResponse.new(response)
 
     @session_id = response_packet.header.session_id
@@ -131,12 +127,6 @@ class Smb2::Client
       ),
     )
 
-    # copy semantics are a pain, dance around it with the reassignment polka
-    header = packet.header
-    header.command_seq = @sequence_number += 1
-    header.session_id = @session_id
-    packet.header = header
-
     ssp_offset = response_packet.security_blob.index("NTLMSSP")
     resp_blob = response_packet.security_blob.slice(ssp_offset .. -1)
 
@@ -145,8 +135,7 @@ class Smb2::Client
     @session_key = type3.session_key
 
     packet.security_blob = gss_type3(type3.serialize)
-    dispatcher.send_packet(packet)
-    response = dispatcher.recv_packet
+    response = send_recv(packet)
     response_packet = Smb2::Packet::SessionSetupResponse.new(response)
 
     response_packet.header.nt_status
@@ -163,8 +152,7 @@ class Smb2::Client
       client_guid: 0,
     )
 
-    dispatcher.send_packet(packet)
-    response = dispatcher.recv_packet
+    response = send_recv(packet)
     response_packet = Smb2::Packet::NegotiateResponse.new(response)
 
     @capabilities  = response_packet.capabilities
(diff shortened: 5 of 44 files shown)


jlee-r7:
Crap, I forgot 2.0.0 doesn't support required keyword args.  :sob: 

@limhoff-r7 Do you think we need to support 2.0.0?

jlee-r7:
Well hell, jruby doesn't support require keyword args either.  That's maybe more of an issue.

limhoff-r7:
Ruby 2.0 added keyword arguments (`def foo(bar: default)`).  Ruby 2.1 added *required* keyword arguments (`def foo(bar:)`).  JRuby 9000 is supposed to be Ruby 2.2 compatible.  We've dropped support for Ruby 2.0 in the rest of the ecosystem, so I'm fine with saying smb2 needs Ruby 2.1+ features, which implies JRuby 9000.

jlee-r7:
This fixes #14

README.md
```diff
@@ -30,17 +30,7 @@ Or install it yourself as:
...
-c = Smb2::Client.new(
-  socket: sock,
-  username:"administrator",
-  password:"P@ssword1",
-  domain:"asdfasdf"
-)
-c.negotiate
-c.authenticate
-```
+{include:Smb2::Client}
```
limhoff-r7: Include directives are YARD-only, so they're just rendered as literal on Github.

limhoff-r7: See https://github.com/jlee-r7/smb2/blob/feature/14/create-file/README.md

README.md
```diff
@@ -60,9 +50,14 @@ neg_response = Smb2::Packet::NegotiateResponse.new(data)
 
 ```
 
+## License
+
+{include:file:LICENSE.txt}
```
limhoff-r7: Include directives are YARD-only, so they're just rendered as literal on Github.  See https://github.com/jlee-r7/smb2/blob/feature/14/create-file/README.md

examples/read_file.rb
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env ruby
+
+$:.push File.expand_path(File.join(File.dirname(__FILE__), '..', 'lib'))
```
limhoff-r7: I'd prefer using `require 'bundler/setup'`

lib/smb2.rb
```diff
@@ -2,10 +2,12 @@
...
 # [[MS-SMB2] Server Mesage Block (SMB) Protocol Versions 2 and 3](https://msdn.microsoft.com/en-us/library/cc246482.aspx)
 module Smb2
-  autoload :VERSION, 'smb2/version'
-  autoload :Packet, 'smb2/packet'
-  autoload :Dispatcher, 'smb2/dispatcher'
   autoload :Client, 'smb2/client'
+  autoload :Dispatcher, 'smb2/dispatcher'
+  autoload :Packet, 'smb2/packet'
+  autoload :Tree, 'smb2/tree'
+  autoload :File, 'smb2/file'
```
limhoff-r7: Please keep the autoloads ordered by the constant name so that it's easier to spot if a constant is missing.

lib/smb2/client.rb
```diff
@@ -1,7 +1,21 @@
 require 'smb2/packet'
```
limhoff-r7: You're using `Kernel.autoload` in `lib/smb2.rb`, so you shouldn't be using explicit `require` here as it will add an unnecessary scan of `$LOAD_FEATURES` when the constants are already loaded.

README.md
jlee-r7: Dammit, I thought I was being clever.  :disappointed:

lib/smb2/client.rb
```diff
@@ -1,7 +1,21 @@
...
+require 'smb2/tree'
 require 'net/ntlm'
```
limhoff-r7: Please separate requires for gem files from project files so it's easier to track the source like so:

```ruby
#
# Standard Library
#

...

#
# Gems
#

...

#
# Project
#

...
```

If you [remove the unneed project requires](https://github.com/rapid7/smb2/pull/28/files#r28818234), this would just degenerate to

```ruby
#
# Gems
#

require 'net/ntlm'
require 'net/ntlm/client'
```

lib/smb2/client.rb
```diff
@@ -1,7 +1,21 @@
...
 require 'net/ntlm/client'
 
+# A client for holding the state of an SMB2 session.
+#
+# ```ruby
```
limhoff-r7: Usage example in YARD should go in [`@example` tags](http://www.rubydoc.info/gems/yard/file/docs/Tags.md#example):

```ruby
# A client for holding the state of an SMB2 session.
#
# @example Authenticating to a SMB2 server with username, password, and domain
#     sock = TCPSocket.new("192.168.100.140", 445)
#     c = Smb2::Client.new(
#       socket: sock,
#       username:"administrator",
#       password:"P@ssword1",
#       domain:"asdfasdf"
#     )
#     c.negotiate
#     c.authenticate
```

lib/smb2/client.rb
```diff
@@ -74,6 +88,8 @@ def initialize(opts = {})
     @password    = opts.fetch(:password).encode("utf-8")
     @domain      = opts[:domain]
     @local_workstation = (opts[:local_workstation] || "")
+
```
limhoff-r7: Since you're requiring required keyword argument support from Ruby 2.1, you can change these `opt.fetch` calls to required keyword arguments.  You can also change the `:domain` and `:local_workstation` to optional keyword arguments:

```ruby
  # @option opts [#send_packet,#recv_packet] :dispatcher
  # @option opts [String] :username UTF-8
  # @option opts [String] :password UTF-8
  # @option opts [String] :domain UTF-8 *(optional)*
  # @option opts [String] :local_workstation UTF-8 *(optional)*
  # @raise KeyError if required options are missing
  def initialize(dispatcher:, domain: nil, local_workstation: '', password:, username:)
    @dispatcher  = dispatcher
    @username    = username.encode("utf-8")
    @password    = password.encode("utf-8")
    @domain      = domain
    @local_workstation = local_workstation

    @tree_ids = []
  end
```

**NOTE: This change can be done in a separate Chore PR since it's not strictly related to adding Read files support**

lib/smb2/client.rb
```diff
@@ -157,10 +160,42 @@ def negotiate
...
+    response_packet = Smb2::Packet::TreeConnectResponse.new(response)
+
+    Smb2::Tree.new(client: self, tree_connect_response: response_packet)
+  end
+
+  # Adjust `request`'s header with an appropriate sequence number and session
+  # id, then send it and wait for a response.
+  #
+  # @return (see Dispatcher::Socket#recv_packet)
+  def send_recv(request)
```
limhoff-r7: Within a given visibility (public, protected, private) please keep methods ordered by name.

lib/smb2/client.rb
```diff
@@ -157,10 +160,42 @@ def negotiate
 
     @state = :negotiated
 
-    @sequence_number = 0
     # XXX do we need the Server GUID?
   end
 
+  # @param tree [String]
```
limhoff-r7: Public methods should have descriptions in their docs.  `@return` is also missing.  Might want to add an `@see` to the Microsoft docs too.

lib/smb2/dispatcher/socket.rb
```diff
@@ -7,7 +9,7 @@ class Smb2::Dispatcher::Socket < Smb2::Dispatcher::Base
   # @param host [String] passed to TCPSocket.new
   # @param port [Fixnum] passed to TCPSocket.new
   def self.connect(host, port)
-    new(TCPSocket.new(host, port))
+    new(::TCPSocket.new(host, port))
```
limhoff-r7: Will the socket class need to be injected for usage in metasploit-framework to support tunnelled sockets?

lib/smb2/packet.rb
```diff
@@ -132,44 +145,56 @@ def self.data_buffer(name, bit_length=16, opts={})
...
+      if !data_buffer_fields.empty?
+        data_buffer_fields.each do |buffer_name|
           @data_buffers[buffer_name] = self.send(buffer_name) || ""
         end
         recalculate
       end
       yield self if block_given?
     end
+
+    if respond_to?(:header) && self.class.respond_to?(:command)
```
limhoff-r7: Is the command ever dynamic on a class?  If not you can do a class-scoped constant lookup with `self.class::COMMAND`.

lib/smb2/packet/close_request.rb
```diff
@@ -9,6 +9,11 @@ class CloseRequest < Smb2::Packet
     unsigned :flags, 16
     unsigned :reserved, 32
     string :file_id, 128
+
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Unnecessary if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/create_request.rb
```diff
@@ -40,5 +44,9 @@ class CreateRequest < Smb2::Packet
 
     rest :buffer
 
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Unnecessary if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/ioctl_request.rb
```diff
@@ -65,5 +65,10 @@ class IoctlRequest < Smb2::Packet
       SMB2_0_IOCTL_IS_FSCTL: 1
     }.freeze
 
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Unnecessary if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/negotiate_request.rb
```diff
@@ -17,6 +17,12 @@ class NegotiateRequest < Smb2::Packet
     # Just 2.02 for now. XXX Update dialect_count if you add anything here
     # XXX :default doesn't do anything at all on rest fields
     rest :dialects #, default: "\x02\x02"
+
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Unnecessary if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/query_info_request.rb
```diff
@@ -35,6 +35,11 @@ class QueryInfoRequest < Smb2::Packet
     string :file_id, 128
 
     rest :buffer
+
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Unnecessary if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/read_request.rb
```diff
@@ -47,13 +47,32 @@ class ReadRequest < Smb2::Packet
...
+    # @note Updating this will *NOT* change {#read_channel_info_length} or
+    #   {#read_channel_info_offset}
+    string :read_channel_info, 8, default: "\x00".force_encoding('binary')
 
     FLAGS = {
       READ_UNBUFFERED: 0x01
     }.freeze
+
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Unnecessary if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/query/standard_information.rb
```diff
@@ -0,0 +1,14 @@
+# [[MS-FSCC] 2.4.38 FileStandardInformation](https://msdn.microsoft.com/en-us/library/cc232088.aspx)
+class Smb2::Packet::Query
+  class StandardInformation < BitStruct
+    default_options endian: 'little'
+
+    unsigned :allocation_size, 64
+    unsigned :end_of_file, 64
+    unsigned :number_of_links, 32
+    unsigned :delete_pending, 8
```
limhoff-r7: They seriously wasted a byte on a boolean for alignment sake?

lib/smb2/packet/session_setup_request.rb
```diff
@@ -36,13 +36,10 @@ class SessionSetupRequest < Smb2::Packet
     }.freeze
     FLAG_NAMES = FLAGS.keys
 
-    def initialize(*args)
-      super
-      new_header = self.header
-      new_header.command = Smb2::COMMANDS[:SESSION_SETUP]
-      self.header = new_header
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Change to `COMMAND = SESSION_SETUP` if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/tree_connect_request.rb
```diff
@@ -4,13 +4,18 @@ class Smb2::Packet
...
     # These two bytes are used in the response, but just padding in the
     # request
-    unsigned :unused, 16
+    unsigned :reserved, 16, default: 0
     data_buffer :tree
 
     rest :buffer
 
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Change to `COMMAND = TREE_CONNECT` if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/packet/query/standard_information.rb
jlee-r7: All over the place.

lib/smb2/packet.rb
jlee-r7: I actually started implementing it that way originally and went to a method instead.  I don't have a good reason why.

lib/smb2/packet/tree_connect_request.rb
```diff
@@ -4,13 +4,18 @@ class Smb2::Packet
   # [Section 2.2.9 SMB2 TREE_CONNECT Request](https://msdn.microsoft.com/en-us/library/cc246567.aspx)
   class TreeConnectRequest < Smb2::Packet
     nest :header, RequestHeader
-    unsigned :struct_size, 16
+    unsigned :struct_size, 16, default: 9
```
limhoff-r7: It's not obvious why this would default at `9` without a derivation from flag until a developer reads link.  Can you add a `@note` with a quote of "StructureSize (2 bytes): The client MUST set this field to 9, indicating the size of the request structure, not including the header. The client MUST set it to this value regardless of how long Buffer[] actually is in the request being sent." like you did for the other must be set field?

lib/smb2/packet/write_request.rb
```diff
@@ -23,5 +23,9 @@ class WriteRequest < Smb2::Packet
 
     rest :buffer
 
+    # @return [Symbol] a key in {Smb2::COMMANDS}
+    def self.command
```
limhoff-r7: Change to `COMMAND = WRITE` if you move to [class-constant lookup](https://github.com/rapid7/smb2/pull/28/files#r28820364)

lib/smb2/tree.rb
```diff
@@ -0,0 +1,51 @@
+
+# A connected tree, as returned by a {Smb2::Packet::TreeRequest}.
+class Smb2::Tree
+
+  attr_accessor :client
```
limhoff-r7: Missing comments for `attr_accessor`s.

lib/smb2/tree.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  attr_accessor :tree_connect_response
+
+  # @param client [Smb::Client]
+  # @param tree_connect_response [Smb::Packet::TreeConnectResponse]
+  def initialize(client:, tree_connect_response:)
+    raise ArgumentError unless tree_connect_response.is_a?(Smb2::Packet::TreeConnectResponse)
```
limhoff-r7: Since you're checking for a specific class I think `TypeError` would be more appropriate.

spec/lib/smb2/packet/close_request_spec.rb
```diff
@@ -1,5 +1,5 @@
+require 'spec_helper'
```
limhoff-r7: If you're using rspec >= 3 you can do the following in `.rspec`:

```
--require spec_helper
```

and never have to `require 'spec_helper'` at the top of your specs again.  I've been doing that as I convert metasploit-* projects to rspec 3.1/3.2.

spec/lib/smb2/packet/close_request_spec.rb
```diff
@@ -1,5 +1,5 @@
...
 require 'smb2'
```
limhoff-r7: You can get rid of this if you just `require 'smb2'` in `spec/spec_helper.rb`.

spec/lib/smb2/packet/create_request_spec.rb
```diff
@@ -17,15 +17,19 @@
...
       expect(packet.struct_size).to eq(57)
       expect(packet.oplock).to eq(0)
+      expect(packet.security_flags).to eq(0)
       expect(packet.impersonation).to eq(2)
-      expect(packet.access_mask).to eq(0)
       expect(packet.create_flags).to eq(0)
 
+      expect(packet.reserved).to eq(0)
+
+      expect(packet.desired_access).to eq(0x0012_019f)
```
limhoff-r7: Could use some local variables and flags combining to explain this value.

spec/lib/smb2/packet/create_request_spec.rb
```diff
@@ -17,15 +17,19 @@
...
       expect(packet.file_attributes).to eq(0)
       # SHARE_DELETE | SHARE_WRITE | SHARE_READ
```
limhoff-r7: Comments like this are also helpful to fix [this](https://github.com/rapid7/smb2/pull/28/files#r28821889)

spec/lib/smb2/packet/read_response_spec.rb
```diff
@@ -21,13 +21,20 @@
       expect(packet.struct_size).to eq(17)
     end
 
-    specify 'data' do
+    specify do
       expect(packet.data_length).to eq(92)
+    end
+    specify do
```
limhoff-r7: Please put blank lines between your `do .. end` blocks like here with your `specify` calls.

spec/lib/smb2/packet/read_response_spec.rb
```diff
@@ -21,13 +21,20 @@
...
+      expect(packet.data.length).to eq(92)
```
limhoff-r7: If `packet.data_length == packet.data.length` is an invariant that should have an explicit example instead of expecting that both individually are the same value, 92.

spec/lib/smb2/packet_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'smb2'
+
+RSpec.describe Smb2::Packet do
+  subject(:packet) do
+    klass = Class.new(described_class)
+    klass.instance_eval do
```
limhoff-r7: You can just pass a block to `Class.new` to eval inside it:

```ruby
klass = Class.new(described_class) do
  nest :header, Smb2::Packet::RequestHeader
      # struct_size is part of the API and must be present
      # 2 bytes for struct_size itself, 2 for data_offset, 4 for data_length
      unsigned :struct_size, 16, default: 2 + 2 + 4
      data_buffer :data, 32
      rest :buffer
end
```

spec/lib/smb2/packet_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      # struct_size is part of the API and must be present
+      # 2 bytes for struct_size itself, 2 for data_offset, 4 for data_length
+      unsigned :struct_size, 16, default: 2 + 2 + 4
+      data_buffer :data, 32
+      rest :buffer
+    end
+    klass.new
+  end
+
+  it { is_expected.to respond_to(:recalculate) }
```
limhoff-r7: Please keep your respond_to expectations ordered by message name so it's easier to spot if one is missing.

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - '2.0'
   - '2.1'
   - '2.2'
-  - 'jruby-19mode'
+  - 'jruby-9000'
```
limhoff-r7: Aw, just missed it

it__s_over_9000_

spec/lib/smb2/packet_spec.rb
jlee-r7: Oh, neat. TIL.

limhoff-r7:
`file.read` returns `nil` with a `file.size` of 739856, but I can read pieces if I give an offset and size: 

```irb
2.1.5 :059 > file.read offset: 0, length: 10
 => "MZ\x90\x00\x03\x00\x00\x00\x04\x00"
```

lib/smb2/dispatcher/socket.rb
jlee-r7: I had envisioned metasploit-framework implementing its own `Dispatcher`.

jlee-r7:
Added `File#read_all` to solve this issue.

PR #30, opened 2015-04-30
Create for writing

### LINKED ISSUE #29: Client: Create file for writing
something like this:
```ruby
tree = client.tree_connect("\\\\hostname\\share")
file = tree.create("file.txt", "wb")
file.write(data)
file.close
```
needs to 
- send a `CreateRequest` with appropriate values
- read the response, store the `file_id` therein
  - raise if error
- send a `SetInfo` request to truncate the file
- read response
  - raise if error
- send a `WriteRequest` with the given data
- read response
  - raise if error

jlee-r7:
Fixes #29 

## Verification

- [x] `bundle console`
- [x] setup 
```ruby
host = ...
dispatcher = Smb2::Dispatcher::Socket.connect(host, 445)
client = Smb2::Client.new(dispatcher: dispatcher, username:"Administrator", password:"notpassword")
client.negotiate
client.authenticate
tree = client.tree_connect("\\\\#{dispatcher.socket.remote_address.ip_address}\\c$")
```

## Writing with seek
- [x] `file = tree.create("yourmom.txt", "w")`
- [x] **Verify** `file.pos` is `0`.
- [x] `file.write("yourmom\n")`
- [x] **Verify** `file.pos` is `8`
- [x] 
```ruby
file.seek 0
file.read
```
- [x] **Verify** `"yourmom\n"` returned.
- [x] 
```ruby
file.write("second line\n")
file.seek 0
file.read
```
- [x] **Verify** `"yourmom\nsecond line\n"` returned
- [x] 
```ruby
file.seek 4
file.write("bob")
file.seek 0
file.read
```
- [x] **Verify** `"yourbob\nsecond line\n"` returned
- [x] `file.close`

## Append puts the pointer at the end
- [x] 
```ruby
file = tree.create("yourmom.txt", "a")
file.write("more stuff\n")
file.seek 0
file.read
```
- [x] **Verify** `"yourbob\nsecond line\nmore stuff\n"` returned

## Write truncates
- [x] `file = tree.create("yourmom.txt", "w")`
- [x] **Verify** `file.pos` is `0`.
- [x] **Verify** `file.size` is `0`.


lib/smb2/client.rb
```diff
@@ -193,17 +190,19 @@ def send_recv(request)
...
+  # Connect to a share
+  #
+  # @param tree [String] Something like "\\\\hostname\\tree"
   # @return [Smb2::Tree]
   def tree_connect(tree)
-    packet = Smb2::Packet::TreeConnectRequest.new do |request|
-      request.tree = tree.encode("utf-16le")
-    end
+    packet = Smb2::Packet::TreeConnectRequest.new(
+      tree: tree.encode("utf-16le")
```
limhoff-r7: Thanks Microsoft! For being consistent and having username and password be utf-8, but paths be utf-16le.

limhoff-r7: I like passing the tree in as a constructor argument over having to use a block.  Good change.

lib/smb2/file.rb
```diff
@@ -4,29 +4,43 @@
...
-  attr_accessor :tree
+  # Some day we'll document all of these in one place
+  STATUS_END_OF_FILE = 0xC0000011
 
   # The server's response from when we {Tree#create created} this File
   #
   # @return [Smb2::Packet::CreateResponse]
   attr_accessor :create_response
 
+  # @return [String]
```
limhoff-r7: This should have a description even if most people will understand want a filename is.  Explain what should go in Smb2::Tree#share vs Smb2::File#filename by showing an example full UNC path being broken up into each piece.

lib/smb2/file.rb
```diff
@@ -4,29 +4,43 @@
...
+  #
+  # @return [Smb2::Tree]
+  attr_accessor :tree
+
   # @param tree [Smb2::Tree] the Tree where this File was opened. See {#tree}
   # @param create_response [Smb2::Packet::CreateResponse] the server's
   #   response from when we {Tree#create created} this File. See
   #   {#create_response}
-  def initialize(tree:, create_response:)
+  def initialize(filename:, tree:, create_response:)
```
limhoff-r7: `@param filename` is missing.

lib/smb2/file.rb
```diff
@@ -4,29 +4,43 @@
...
+    self.filename = filename
     self.tree = tree
     self.create_response = create_response
+    self.pos = 0
```
limhoff-r7: Thanks for using the attr_writers inside initialize instead of setting instance variables directly.  I prefer this style as it makes patching the code easier.

lib/smb2/file.rb
```diff
@@ -34,21 +48,28 @@ def initialize(tree:, create_response:)
   #
   # @return [Smb2::Packet::CloseResponse]
   def close
-    packet = Smb2::Packet::CloseRequest.new do |request|
-      request.file_id = self.create_response.file_id
-    end
+    packet = Smb2::Packet::CloseRequest.new(
```
limhoff-r7: I like the change to constructor arguments instead of blocks.

lib/smb2/file.rb
```diff
@@ -34,21 +48,28 @@ def initialize(tree:, create_response:)
...
     response = tree.send_recv(packet)
 
     Smb2::Packet::CloseResponse.new(response)
   end
 
+  # Whether the current read/write pointer is at the end of the file
+  #
+  # @return [Boolean]
+  def eof?
+    (@last_read_response && @last_read_response.header.nt_status == STATUS_END_OF_FILE) || pos == size
```
limhoff-r7: I'd prefer a private attr_reader usage than direct instance variable access here.

lib/smb2/file.rb
```diff
@@ -34,21 +48,28 @@ def initialize(tree:, create_response:)
...
+  end
+
   # @return [String]
   def inspect
-    "#<Smb2::File file-id=#{create_response.file_id.unpack("H*").first} >"
+    "#<Smb2::File:#{filename} file-id=#{create_response.file_id.unpack("H*").first} >"
```
limhoff-r7: This is a useful change, but wouldn't it be more useful to have the full UNC path since this won't include which share you're attached to since it doesn't include the tree.share?

lib/smb2/file.rb
```diff
@@ -34,21 +48,28 @@ def initialize(tree:, create_response:)
...
   end
 
-  # Call {#read_chunk} repeatedly until we get everything
+  # Call {#read_chunk} repeatedly until we get `length` bytes or hit eof.
```
limhoff-r7: Link to `#eof?` when you say `hit eof`, so either `{#eof? hit eof}` or `hit {#eof? eof}` since it expands eof out to "end of file".  EOF may be a new abbreviation for the reader.

lib/smb2/file.rb
```diff
@@ -58,23 +79,23 @@ def inspect
   # @param length [Fixnum,nil] number of bytes to read. If this is nil, read
   #   the whole file, i.e. {#size} bytes.
   # @return [String] full contents of the remote file
-  def read(offset: 0, length: nil)
+  def read(length = nil, offset: self.pos)
```
limhoff-r7: I'm not a huge fan of this mix of positional arguments with defaults and keyword arguments.  Was this so that our API was closer to the Ruby File API?

lib/smb2/file.rb
```diff
@@ -58,23 +79,23 @@ def inspect
...
 
-    # Starting from `offset`, up to `length` bytes after `offset`, counting by
-    # maximum chunk size
-    (offset...(offset+length)).step(max) do |step|
+    while data.length < length && !eof?
       # when we are close to the end, we need to read fewer then max bytes
       len = [ max, length - data.length ].min
-      data << read_chunk(offset: step, length: len)
+      response_packet = read_chunk(offset: pos, length: len)
+      data << response_packet.data
```
limhoff-r7: Is `response_packet.data` guaranteed not to be `nil` here?

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
+      minimum_count: 0
+    )
 
     response = tree.send_recv(packet)
 
     response_packet = Smb2::Packet::ReadResponse.new(response)
     @last_read_response = response_packet
 
-    response_packet.data
+    seek(response_packet.data_length, IO::SEEK_CUR)
```
limhoff-r7: Good reuse of the Ruby stdlib constants.

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
+
+    response_packet
+  end
+
+  # Seeks to a given `offset` in the stream according to the value of
+  # `whence`:
+  #
+  # |`whence`||
+  # |----------------------|--------------------------------------------------
+  # | :CUR or IO::SEEK_CUR | Seeks to `amount` plus current position
```
limhoff-r7: The constants have anchors on ruby-doc.org if you use developer tools, so you can link to them with
```
{http://ruby-doc.org/core-2.2.2/IO.html#SEEK_CUR `IO::SEEK_CUR`}
```
, etc

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
+  # | :END or IO::SEEK_END | Seeks to `amount` plus end of stream (you probably want a negative value for `amount`)
+  # | :SET or IO::SEEK_SET | Seeks to the absolute location given by `amount`
+  #
+  # @see http://ruby-doc.org/core-2.2.2/IO.html#method-i-seek
+  # @param amount [Fixnum]
+  # @param whence [Symbol]
+  def seek(amount, whence=IO::SEEK_SET)
```
limhoff-r7: Good use of using compatible API with Ruby IO.

limhoff-r7: You should add a `@return` tag.  Either `@return [Integer]` if returning `@pos` is part of the API or `@return [void]` if callers should not depend on the return value.  Right YARD documents it as the default `@return [Object]`:
screen shot 2015-05-06 at 1 15 45 pm

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
   # @return [Fixnum]
   def size
-    packet = Smb2::Packet::QueryInfoRequest.new do |request|
-      request.info_type = Smb2::Packet::QUERY_INFO_TYPES[:FILE]
-      request.file_info_class = Smb2::Packet::FILE_INFORMATION_CLASSES[:FileStandardInformation]
-      request.output_buffer_length = 40
-      request.input_buffer_length = 0
-      request.file_id = self.create_response.file_id
-    end
+    packet = Smb2::Packet::QueryInfoRequest.new(
```
limhoff-r7: Once again, good change away from needing the block.

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
+      info_type: Smb2::Packet::QUERY_INFO_TYPES[:FILE],
+      file_info_class: Smb2::Packet::FILE_INFORMATION_CLASSES[:FileStandardInformation],
+      output_buffer_length: 40,
```
limhoff-r7: Why `40`?  Can you put this in a constant with a descriptive comment explain why?

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
+  # |----------------------|--------------------------------------------------
+  # | :CUR or IO::SEEK_CUR | Seeks to `amount` plus current position
+  # | :END or IO::SEEK_END | Seeks to `amount` plus end of stream (you probably want a negative value for `amount`)
+  # | :SET or IO::SEEK_SET | Seeks to the absolute location given by `amount`
+  #
+  # @see http://ruby-doc.org/core-2.2.2/IO.html#method-i-seek
+  # @param amount [Fixnum]
+  # @param whence [Symbol]
+  def seek(amount, whence=IO::SEEK_SET)
+    @pos = case whence
```
limhoff-r7: Could you change this to `self.pos =` to avoid directly accessing the instance variable since using the writer can be intercepted?

lib/smb2/file.rb
```diff
@@ -84,34 +105,59 @@ def read(offset: 0, length: nil)
...
+           when :CUR, IO::SEEK_CUR
+             @pos + amount
```
limhoff-r7: Similarly, change this to `pos + amount` for easier intercept.

lib/smb2/file.rb
```diff
@@ -119,4 +165,51 @@ def size
...
+    max = tree.client.max_write_size
+    bytes_written = 0
+    seek(offset)
+
+    while data.length > bytes_written
+      data_chunk = data.slice(bytes_written, max)
+      response_packet = write_chunk(data_chunk, offset: pos)
+
+      # @todo raise instead?
+      break if response_packet.header.nt_status != 0
```
limhoff-r7: You could do `break unless response.header.nt_status == 0` since you're using `!=`.

lib/smb2/file.rb
```diff
@@ -119,4 +165,51 @@ def size
     info.end_of_file
   end
 
+  # Write the entire contents of `data`, starting at `offset` bytes from the
+  # beginning of the file.
+  #
+  # @param data [String] what to write
+  # @param offset [Fixnum] where in the file to start writing
+  # @return [Fixnum] number of bytes written. This may be less than the length
```
limhoff-r7: Is there anyway for the caller to get the error?  If so, that should be documented.

lib/smb2/file.rb
```diff
@@ -119,4 +165,51 @@ def size
...
+  #   of `data` if there was an error.
+  def write(data, offset: self.pos)
+    max = tree.client.max_write_size
+    bytes_written = 0
+    seek(offset)
+
+    while data.length > bytes_written
+      data_chunk = data.slice(bytes_written, max)
```
limhoff-r7: Do we have any idea on the efficiency of slice when chunking the data?  Would `each_slice` be more efficient?

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
-  QUERY_INFO_TYPES = {
-    # SMB2_0_INFO_FILE
-    FILE: 0x01,
-    # SMB2_0_INFO_FILESYSTEM
-    FILESYSTEM: 0x02,
-    # SMB2_0_INFO_SECURITY
-    SECURITY: 0x03,
-    # SMB2_0_INFO_QUOTA
-    QUOTA: 0x04
+  # Used in {CreateRequest#disposition}
```
limhoff-r7: Could you add `@note Ordered by key value` so that I don't think they're out of order and so that other developers understand where to insert new values?

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
+    FILE_OPEN_IF: 0x00000003,
+    # Overwrite the file if it already exists; otherwise, fail the operation. MUST
+    # NOT be used for a printer object.
+    FILE_OVERWRITE: 0x00000004,
+    # Overwrite the file if it already exists; otherwise, create the file.
+    # This value SHOULD NOT be used for a printer object.<32>
+    FILE_OVERWRITE_IF: 0x00000005,
+  }.freeze
+
+  # Used in {CreateRequest#create_options}
```
limhoff-r7: Could you add `@note Ordered by key value`?

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
+  #
+  # https://msdn.microsoft.com/en-us/library/cc246502.aspx
+  CREATE_OPTIONS = {
+    FILE_DIRECTORY_FILE: 0x0000_0001,
```
limhoff-r7: Would it make the value derivation clearer if you did explicit shifts in the code or is it enough to just have the link to the table?

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
+    # NOT be used for a printer object.
+    FILE_OVERWRITE: 0x00000004,
+    # Overwrite the file if it already exists; otherwise, create the file.
+    # This value SHOULD NOT be used for a printer object.<32>
+    FILE_OVERWRITE_IF: 0x00000005,
+  }.freeze
+
+  # Used in {CreateRequest#create_options}
+  #
+  # https://msdn.microsoft.com/en-us/library/cc246502.aspx
```
limhoff-r7: Did you mean to have `@see` before this link?

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
+    FILE_OPEN_REMOTE_INSTANCE: 0x0000_0400,
+    FILE_OPEN_REQUIRING_OPLOCK: 0x0001_0000,
+    FILE_DISALLOW_EXCLUSIVE: 0x0002_0000,
+    FILE_RESERVE_OPFILTER: 0x0010_0000,
+    FILE_OPEN_REPARSE_POINT: 0x0020_0000,
+    FILE_OPEN_NO_RECALL: 0x0040_0000,
+    FILE_OPEN_FOR_FREE_SPACE_QUERY: 0x0080_0000,
+  }.freeze
+
+  # Used in {CreateRequest#desired_access} when opening a file, pipe, printer.
```
limhoff-r7: Please add `@note Ordered by key value`.

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
+  #
+  # From the documentation ([2.2.13.1.1 File_Pipe_Printer_Access_Mask](https://msdn.microsoft.com/en-us/library/cc246802.aspx)):
+  #
+  # > The SMB2 Access Mask Encoding in SMB2 is a 4-byte bit field value that
+  #   contains an array of flags. An access mask can specify access for one of
+  #   two basic groups: either for a file, pipe, or printer (specified in
+  #   section 2.2.13.1.1) or for a directory (specified in section 2.2.13.1.2).
+  #
+  # @see DIRECTORY_ACCESS_MASK
+  FILE_ACCESS_MASK = {
```
limhoff-r7: Similar to https://github.com/rapid7/smb2/pull/30/files#r29786993, would showing the shifts make the derivation clearer?

lib/smb2/packet.rb
```diff
@@ -47,17 +47,134 @@ class InvalidFlagError < StandardError; end
...
+    GENERIC_WRITE: 0x40000000,
+    # This value indicates a request for the following combination of access
+    # flags listed above: FILE_READ_DATA| FILE_READ_ATTRIBUTES| FILE_READ_EA|
+    # SYNCHRONIZE| READ_CONTROL.
+    GENERIC_READ: 0x80000000,
+  }.freeze
+
+  # [2.2.13.1.2 Directory_Access_Mask](https://msdn.microsoft.com/en-us/library/cc246801.aspx)
+  # @todo
+  DIRECTORY_ACCESS_MASK = {
```
limhoff-r7: Did you mean for this to be empty?

lib/smb2/packet.rb
```diff
@@ -174,13 +304,13 @@ def self.data_buffer(name, bit_length=16, opts={})
   def initialize(*args)
     @data_buffers = {}
     super do
+      yield self if block_given?
```
limhoff-r7: You should document the block using `@yield`, `@yieldparam` and `@yieldreturn`.  Since you're not capturing the return value from `yield` you should do `@yieldreturn [void]` to indicate that nothing is done with the return value from the block.

lib/smb2/tree.rb
```diff
@@ -14,12 +19,13 @@ class Smb2::Tree
 
   # @param client [Smb::Client]
   # @param tree_connect_response [Smb::Packet::TreeConnectResponse]
-  def initialize(client:, tree_connect_response:)
+  def initialize(client:, share:, tree_connect_response:)
```
limhoff-r7: Missing `@param` for `share`.

lib/smb2/tree.rb
```diff
@@ -28,23 +34,68 @@ def initialize(client:, tree_connect_response:)
...
-    packet = Smb2::Packet::CreateRequest.new do |request|
-      request.filename = filename.encode("utf-16le")
-      # @todo document all these flags. value copied from smbclient traffic
-      request.desired_access = 0x0012_0089
-      request.impersonation = 2
-      request.share_access = 3  # SHARE_WRITE | SHARE_READ
-      request.disposition = 1   # Open (if file exists open it, else fail)
-      request.create_options = 0x40 # non-directory
-    end
+  def create(filename, mode = "r+")
```
limhoff-r7: Because `create` has different `return` based on whether there is a block given or not you should use `@overload` to document the behavior with and without a block separately.  In addition to `@yield`, you should use `@yieldparam` and `@yieldreturn` since the value returned from the block is captured.  The format for `@yield` in relation to `@yieldparam` is `@yield [param1, param1]` and then those param names are used in `@yieldparam param1 [Type] description`.

lib/smb2/tree.rb
```diff
@@ -28,23 +34,68 @@ def initialize(client:, tree_connect_response:)
...
+    if block_given?
+      value = yield file
+      file.close
+      value
+    else
+      file
+    end
+  end
+
+  def delete(filename)
```
limhoff-r7: You didn't have mean test delete.  Should I have or is it not complete?

lib/smb2/tree.rb
```diff
@@ -28,23 +34,68 @@ def initialize(client:, tree_connect_response:)
...
+    packet = Smb2::Packet::CreateRequest.new(
+      filename: filename.encode("utf-16le"),
+      desired_access: Smb2::Packet::FILE_ACCESS_MASK[:FILE_DELETE],
+      impersonation: 2,
```
limhoff-r7: `2` should be in a constant explain what it means here.

lib/smb2/tree.rb
```diff
@@ -28,23 +34,68 @@ def initialize(client:, tree_connect_response:)
...
+      share_access: 7, # SHARE_DELETE | SHARE_WRITE | SHARE_READ
```
limhoff-r7: Please define the constants in the comment and `|` together instead of hard-coding `7` as a magic number.

lib/smb2/file.rb
jlee-r7: Good point.

lib/smb2/file.rb
jlee-r7: Yes, it will always be a (possibly empty) string.

lib/smb2/file.rb
jlee-r7: Yes. It's a little weird though, since [IO#read](http://ruby-doc.org/core-2.2.2/IO.html#method-i-read) just uses the current position pointer; it doesn't take an offset at all. I thought about just changing this to mirror that exactly, but it's a weird API because of the bizarre and very un-Ruby-like `outbuf` argument. What do you think?

lib/smb2/file.rb
jlee-r7: I'm a bit disinclined to do so because the lines in this table are already quite long. Since it's a table, I can't wrap the lines without breaking it and making them longer will sacrifice readability in the code.

lib/smb2/file.rb
jlee-r7: No, unfortunately. I thought about having a `last_write_response` accessor similar to `last_read_response` to solve this problem.

lib/smb2/file.rb
jlee-r7: `#each_slice` is no longer a method on Strings since they are not Enumerable after 1.9

lib/smb2/packet.rb
jlee-r7: Yes, hence the `@todo`. This will be needed for directory listing later.

lib/smb2/packet.rb
jlee-r7: I don't think so. They are used as a bitmask, but the value itself is unimportant, as are the shift values. In fact, I think it's easier to verify correctness with the documentation as is.

lib/smb2/file.rb
limhoff-r7: If you don't want to link in the table, how about adding individual `@see` links to each one then?

lib/smb2/file.rb
limhoff-r7: I assume `outbuf` is some String reuse optimization thing.  I'm fine with not supporting that if we can determine the actual need for it since some of these lower-level Ruby APIs are more C-ish than Ruby-ish.

The only compromise I could see is doing `*args` and supporting overloading, so it would support `read()`, `read(length)` or `def read(offset: self.pos, length: nil)`.  You could use YARD's `@overload` to document each of these calling conventions, but it makes unpacking the argument more complicated internally.

jlee-r7: Yeah, that seems pretty clunky, and would definitely make the method overly complex. What about abandoning the offset and only allowing `read()` (to get everything from current `pos` to end), and `read(length)` to get a certain number of bytes starting from `pos`?

jlee-r7: it's pretty easy to just `seek(offset)` first if you want an offset.

limhoff-r7: I'm OK with doing `read()` and `read(length)`.  That avoids a method that does two things (seek + read).

lib/smb2/file.rb
```diff
@@ -34,47 +62,53 @@ def initialize(tree:, create_response:)
...
   # @note Beware of ballooning memory usage
   # @note Calling this on a pipe without a `length` is probably a really bad
   #   idea
   #
-  # @param offset [Fixnum] offset from the beginning of the file (*default*: 0)
   # @param length [Fixnum,nil] number of bytes to read. If this is nil, read
   #   the whole file, i.e. {#size} bytes.
   # @return [String] full contents of the remote file
-  def read(offset: 0, length: nil)
+  def read(length = nil)
```
trosen-r7: This change is causing a spec failure due to differences in method arity between here and what the spec expects to use...

PR #52, opened 2015-07-13
Packet factory

### LINKED ISSUE #10: Factory for creating proper Packet types from wire data
For parsing off the wire, we currently have to know the packet type in advance.

It might be better to have a factory for creating the correct type of `Packet` based on data, perhaps with an API like this:
```ruby
pkt = Smb2::Packet.create(data) 
# => #<Smb2::Packet::TreeConnectResponse ... >
```

This would allow `Dispatcher`s to do the parsing without having to know what to expect.

We need a way to grab a packet header from the wire, read the correct number of bytes to get the whole thing, then instantiate the correct Request/Response packet object.

This will require adding a method to Smb2::Packet for determining the full size of header + data buffers.

### LINKED ISSUE #11: Smb2::Packet should be a Module
with packet types inheriting from `Smb2::Packet::Generic` or the like.

jlee-r7:
Makes response parsing the dispatcher's job, instead of making everything that sends a request have to know about what kind of response it should receive.

Fixes #10. Incidentally fixes #11.

### Make sure all the stuff that used to work still does
- [x] Follow [verification steps for writing files](https://github.com/rapid7/smb2/pull/30)
- [x] Follow [verification steps for reading files](https://github.com/rapid7/smb2/pull/28)
- [ ] Follow [verification steps for signed packets](https://github.com/rapid7/smb2/pull/44)


jlee-r7:
Conflict is in `version.rb`, which should be fixed as part of merge anyway.

trosen-r7:
@jlee-r7 this appears to fail the file create use case linked above:

```
 :006 > tree = client.tree_connect("\\\\#{dispatcher.socket.remote_address.ip_address}\\c")
 => #<Smb2::Tree \\172.16.251.175\c >
 :007 > file = tree.create("yourmom.txt", "w")
 => #<Smb2::File:\\172.16.251.175\c\yourmom.txt file-id=00000000000000000000000000000000 >
 :008 > file.pos
 => 0
 :009 > file.write("yourmom\n")
 => 0
 :010 > file.pos
 => 0
```

*Note:* I'm using a Win7 VM w/ admin creds. I had to open c for sharing b/c c$ was unavailable. Please advise if there are other setup steps for this env.

jlee-r7:
Weird, C$ should always be available. Check the `file.create_response.nt_status` to see what the problem was. Also, we need better and more obvious error handling.

trosen-r7:
We can roll @dmaloney-r7 new errors gem in.

Got verification figured out.

PR #58, opened 2015-07-28
Chore/msp 13039/meaningful exceptions

trevrosen:
## Verification
- [ ] VERIFY ALL PASS: `rake spec`

Note: I've made an `Error` module that I expect to later break up if/when we create classes/modules for more things in a specific domain. For instance the `NetBiosSessionService` could later be moved into an area of code concerned w/ NetBios in general. For now, this gives us a central place to locate/document exception classes.

### DIFF
--- a/lib/ruby_smb/dispatcher/socket.rb
+++ b/lib/ruby_smb/dispatcher/socket.rb
@@ -30,13 +30,16 @@ class RubySMB::Dispatcher::Socket < Smb2::Dispatcher::Base
     nil
   end
 
+  # Read a packet off the wire and parse it into a string
+  # Throw Error::NetBiosSessionService if there's an error reading the first 4 bytes,
+  # which are assumed to be the NetBiosSessionService header.
   # @return [String]
   # @todo should return Smb2::Packet
   def recv_packet
     IO.select([@socket])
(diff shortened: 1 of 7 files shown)


lib/ruby_smb/dispatcher/socket.rb
```diff
@@ -30,13 +30,16 @@ def send_packet(packet)
...
+  # which are assumed to be the NetBiosSessionService header.
   # @return [String]
   # @todo should return Smb2::Packet
   def recv_packet
     IO.select([@socket])
-    nbss_header = @socket.read(4)
+    nbss_header = @socket.read(4) # Length of NBSS header. TODO: remove to a constant
     if nbss_header.nil?
-      raise "omg"
+      raise RubySmb::Error::NetBiosSessionService
```
dmaloney-r7: Can we please include a message with this exception like "NBSS Header missing" or something similar?

jlee-r7:
The new exceptions should inherit from `StandardError` instead of `Exception` so a bare `rescue` will catch them.

trevrosen:
Fair point, though we should avoid bare rescues wherever possible. Made the change on master.

PR #63, opened 2015-07-29
Feature/msp 12908/smb parameter block

dmaloney-r7:
Adds the SMB_Parameters block for SMB1 packets

VERIFICATION STEPS
- [ ] `rake spec`
- [ ] VERIFY all specs pass
- [ ] VERIFY that BitStruct Structure matches protocol spec as defined at https://msdn.microsoft.com/en-us/library/ee442058.aspx


lib/ruby_smb/smb1/packet/smb_parameter_block.rb
```diff
@@ -0,0 +1,18 @@
...
+    module Packet
+      # This class represents an SMB_Parameters block structure for SMB1 packets.
+      # [Section 2.2.3.2 Parameter Block](https://msdn.microsoft.com/en-us/library/ee442058.aspx)
+      class SmbParameterBlock < BitStruct
+        unsigned :word_count, 8
+        rest :words
+
+        def words=(value)
+          raise ArgumentError, "value must be a binary string" unless value.kind_of? String
+          self[1,(1 + value.length)] = value.force_encoding('binary')
```
jlee-r7: this line seems like it should just be `super`. am I missing something?

dmaloney-r7: yes, super does not work like that in BitStruct because of the way the attrs are added.

lib/ruby_smb/smb1/packet.rb
```diff
@@ -0,0 +1,8 @@
+require 'bit-struct'
```
trosen-r7: Move this require into `lib/ruby_smb.rb`

While you're at it, do the same with the requires in `client.rb`

Let's make sure that all requires are in that central location other than stuff from stdlib.

spec/spec_helper.rb
```diff
@@ -19,6 +19,7 @@
 end
 
 require 'ruby_smb/smb2'
+require 'ruby_smb/smb1'
```
trosen-r7: These shouldn't be necessary. Put all requires in to the top-level file (`lib/ruby_smb.rb`) and then you can just `require 'ruby_smb'`

dmaloney-r7:
@trosen-r7 fixed

PR #64, opened 2015-07-29
Feature/msp 12907/smb header

dmaloney-r7:
Adds the basic SMB Header for SMB1 Packets

VERIFICATION STEPS
- [ ] `rake spec`
- [ ] VERIFY all specs pass
- [ ] VERIFY BitStruct structure matches MSDN docs https://msdn.microsoft.com/en-us/library/cc246254.aspx
- [ ] Remove PRERELEASE tag

### DIFF
--- a/spec/lib/ruby_smb/smb1/packet/smb_header_spec.rb
+++ b/spec/lib/ruby_smb/smb1/packet/smb_header_spec.rb
@@ -16,4 +16,88 @@ RSpec.describe RubySMB::Smb1::Packet::SmbHeader do
   it { is_expected.to respond_to :pid_low }
   it { is_expected.to respond_to :uid }
   it { is_expected.to respond_to :mid }
+
+  describe 'protocol' do
+    it 'should be a 32-bit field per the SMB spec' do
+      protocol_size_field = header.fields.detect { |f| f.display_name == :protocol }
+      expect(protocol_size_field.length).to eq 32
+    end
+  end
+
+  describe 'command' do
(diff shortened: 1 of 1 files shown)


lib/ruby_smb/smb1.rb
```diff
@@ -0,0 +1,8 @@
+# This module adds the namespace for version 1 of the SMB Protocol
+# as defined in [MS-SMB](https://msdn.microsoft.com/en-us/library/cc246231.aspx)
+module RubySMB::Smb1
+  autoload :Packet, 'ruby_smb/smb1/packet'
+
+  # Protocol ID value. Translates to \xFFSMB
+  SMB_PROTOCOL_ID = 0xFF534D42
```
trosen-r7: Thanks for capturing this in a constant!

spec/lib/ruby_smb/smb1/packet/smb_header_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'spec_helper'
+
+RSpec.describe RubySMB::Smb1::Packet::SmbHeader do
+
+  subject(:header) { described_class.new }
+
+  it { is_expected.to respond_to :protocol }
```
trosen-r7: Make these specs cover the hardcoded int values instead of simply testing that the attr is in place. See [echo_request_spec.rb](https://github.com/rapid7/ruby_smb/blob/master/spec/lib/ruby_smb/smb2/packet/echo_request_spec.rb#L14) for an example.

PR #70, opened 2015-07-31
Feature/msp 13043/socket mocks

trevrosen:
* Adds specs for Socket#recv_packet and Socket::connect
* Socket#socket is now Socket#tcp_socket
* Clean up extraneous requires in specs -- lib code and `require 'spec_helper'`

## Verification
- [x] Verify all specs pass `rake spec`
- [x] Verify that there are no more instances of 'require "spec_helper": `ack spec_helper`
- [x] Verify that the specs show in long-form documentation format
- [x] Verify smoke test works against a Windows box: `ruby examples/smoke_test.rb <YOUR IP AND CREDS>`


trevrosen:
The coverage decrease flag from Coveralls is troubling here given that I've added new specs in this PR...

PR #91, opened 2017-05-17
remove .name from raise

rwhitcroft:
Error here: https://pastebin.com/ebpCQexd

Vanilla run of the module, immediately raises exception. Not sure this is the right fix, just getting the ball rolling.


busterb:
Thanks @rwhitcroft , I proposed an alternate fix, could you verify that it does the right thing? https://github.com/rwhitcroft/ruby_smb/pull/1

dmaloney-r7:
This is not a fix, and in fact actually breaks the code.

bwatters-r7:
@rwhitcroft I think the framework PR https://github.com/rapid7/metasploit-framework/pull/8412 fixed this problem.  Please let us know if it does not.

PR #94, opened 2017-08-31
Add NT_CREATE_ANDX Request and Response

chrisdlf:
Adding the NT_CREATE_ANDX request and response messages and specs
Documentation:
- [2.2.4.64.2 Response](https://msdn.microsoft.com/en-us/library/ee441612.aspx)
- [2.2.4.9.2 Server Response Extensions](https://msdn.microsoft.com/en-us/library/cc246334.aspx)
# Verification Steps
- [x] `bundle install`
## `rake spec`
- [x] `rake spec`
- [x] VERIFY no failures


lib/ruby_smb/smb1/packet/nt_create_andx_request.rb
```diff
@@ -0,0 +1,58 @@
...
+      # [2.2.4.9.1 Client Request Extensions](https://msdn.microsoft.com/en-us/library/cc246332.aspx)
+      class NtCreateAndxRequest < RubySMB::GenericPacket
+
+        # A SMB1 Parameter Block as defined by the {NtCreateAndxRequest}
+        class ParameterBlock < RubySMB::SMB1::ParameterBlock
+          endian :little
+          struct :words, onlyif: -> { word_count.nonzero? } do
+            uint8                              :andx_command,        label: 'AndX Command',       initial_value: 0xFF
+            uint8                              :andx_reserved,       label: 'AndX Reserved',      initial_value: 0x00
+            uint16                             :andx_offset,         label: 'AndX Offset',        initial_value: 0x0000
```
chrisdlf: According to the documentation:

> This field MUST be set to the offset in bytes from the start of the SMB Header (section 2.2.3.1) to the start of the WordCount field in the next SMB command in this packet. This field is valid only if the AndXCommand field is not set to 0xFF. If AndXCommand is 0xFF, this field MUST be ignored by the server.

This logic has not been implemented yet. Do you think it would be better to add this here (in a "value:" attribute for example), or in the client when SMB commands are chained (Andx Messages)?

lib/ruby_smb/smb1/bit_field/smb_ext_file_attributes.rb
```diff
@@ -2,33 +2,30 @@ module RubySMB
...
-        bit8  :reserved5,         label: 'Reserved Space'
-        # Byte Boundary
-        bit1  :write_through,     label: 'Write through'
-        bit1  :reserved6,         label: 'Reserved Space'
-        bit1  :no_buffering,      label: 'Do not Buffer'
-        bit1  :random_access,     label: 'Random Access'
-        bit1  :sequential_scan,   label: 'Sequential Access'
-        bit1  :delete_on_close,   label: 'Delete on close'
-        bit1  :backup_semantics,  label: 'Backup Semantics'
-        bit1  :posix_semantics,   label: 'POSIX Semantics'
```
chrisdlf: I updated this bit-field according to [2.2.1.2.1 Extended File Attribute (SMB_EXT_FILE_ATTR) Extensions](https://msdn.microsoft.com/en-us/library/cc246322.aspx). Please, let me know if you think we should keep the fields defined in CIFS and removed in SMB.

lib/ruby_smb/smb1/bit_field/create_options.rb
```diff
@@ -27,12 +28,12 @@ class CreateOptions < BinData::Record
         # Byte Boundary
         bit1    :open_for_free_space_query,   label: 'Open for Free Space Query'
         bit1    :open_no_recall,              label: 'Open No Recall'
-        bit1    :reserved2,                   label: 'Reserved Space'
+        bit1    :open_reparse_point,          label: 'Open Reparse Point'
```
chrisdlf: I added the FILE_OPEN_REPARSE_POINT field according to [2.2.4.9.1 Client Request Extensions] (https://msdn.microsoft.com/en-us/library/cc246332.aspx).

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
```diff
@@ -0,0 +1,74 @@
+module RubySMB
+  module SMB1
+    module Packet
+
+      # A SMB1 SMB_COM_NT_CREATE_ANDX Response Packet as defined in
+      # [2.2.4.64.2 Response](https://msdn.microsoft.com/en-us/library/ee441612.aspx) and
+      # [2.2.4.9.2 Server Response Extensions](https://msdn.microsoft.com/en-us/library/cc246334.aspx)
```
chrisdlf: According to the documentation ( [2.2.4.9.2 Server Response Extensions] (https://msdn.microsoft.com/en-us/library/cc246334.aspx) ), word_count should be 0x2A, but there is this comment:

https://msdn.microsoft.com/en-us/library/cc246806.aspx#Appendix_A_49
> Windows-based SMB servers send 50 (0x32) words in the extended response although they set the WordCount field to 0x2A

I am confused, is it something we should care about?

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
```diff
@@ -0,0 +1,74 @@
...
+      class NtCreateAndxResponse < RubySMB::GenericPacket
+
+        # A SMB1 Parameter Block as defined by the {NtCreateAndxResponse}
+        class ParameterBlock < RubySMB::SMB1::ParameterBlock
+          endian :little
+
+          struct :words, onlyif: -> { word_count.nonzero? } do
```
chrisdlf: I added the #words structure to be able to read the response message when the server returns an error. In this case, word_count is zero and no parameter block data is returned. Without this, BinData breaks when reading such error response.

lib/ruby_smb/smb1/packet/nt_create_andx_request.rb
```diff
@@ -0,0 +1,58 @@
...
+
+      # A SMB1 SMB_COM_NT_CREATE_ANDX Request Packet as defined in
+      # [2.2.4.64.1 Request](https://msdn.microsoft.com/en-us/library/ee442175.aspx) and
+      # [2.2.4.9.1 Client Request Extensions](https://msdn.microsoft.com/en-us/library/cc246332.aspx)
+      class NtCreateAndxRequest < RubySMB::GenericPacket
+
+        # A SMB1 Parameter Block as defined by the {NtCreateAndxRequest}
+        class ParameterBlock < RubySMB::SMB1::ParameterBlock
+          endian :little
+          struct :words, onlyif: -> { word_count.nonzero? } do
```
chrisdlf: Please see my comment here: https://github.com/rapid7/ruby_smb/pull/94/files#r136414207

lib/ruby_smb/smb1/packet/nt_create_andx_request.rb
```diff
@@ -0,0 +1,58 @@
...
+            create_options                     :create_options,      label: 'Create Options'
+            # The following constants are defined in RubySMB::ImpersonationLevels
+            uint32                             :impersonation_level, label: 'Impersonation Level'
+            security_flags                     :security_flags,      label: 'Security Flags'
+          end
+        end
+
+        # Represents the specific layout of the DataBlock for a {NtCreateAndxRequest} Packet.
+        class DataBlock < RubySMB::SMB1::DataBlock
+          struct :bytes, onlyif: -> { byte_count.nonzero? } do
```
chrisdlf: I added the bytes structure to follow the same logic I explained here: https://github.com/rapid7/ruby_smb/pull/94/files#r136414207

lib/ruby_smb/smb1/bit_field/nt_create_andx_flags.rb
```diff
@@ -0,0 +1,22 @@
+module RubySMB
+  module SMB1
+    module BitField
+      # The Flags bit-field for an SMB1 SMB_COM_NT_CREATE_ANDX Request as defined in
+      # [2.2.4.64.1 Request](https://msdn.microsoft.com/en-us/library/ee442175.aspx)
+      # [2.2.4.9.1 Client Request Extensions](https://msdn.microsoft.com/en-us/library/cc246332.aspx)
+      class NtCreateAndxFlags < BinData::Record
```
dmaloney-r7: If this Bitfield is only ever used in this one packet, you could also just embed this in the packet asa nested struct like so:
https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb1/packet/nt_trans/create_request.rb#L16-L26

lib/ruby_smb/smb1/packet/nt_create_andx_request.rb
dmaloney-r7: There is already an andx_block field defined at https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb1/andx_block.rb that handles everything for the AndX

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
dmaloney-r7: Hrm, i'm not sure how i feel about this. I was just using exception handling an the EmtpyPacket class in SMb1 to deal with this before. This seems like a more convenient solution in some respects, however, it adds another layer of nesting to the structure, and in SMB1 these are already getting pretty nested

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
dmaloney-r7: I doubt it, since this is a response we can probably just ignore this as long as we are parsing the response correctly

spec/lib/ruby_smb/smb1/bit_field/file_status_flags_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  {
+    :reparse_tag    => 0x0004,
+    :no_substreams  => 0x0002,
+    :no_eas         => 0x0001
+  }.each do |field, bitmask|
+    describe "##{field.to_s}" do
+      it 'is a 1-bit flag' do
+        expect(options.send(field)).to be_a BinData::Bit1
+      end
```
dmaloney-r7: please do not ever do meta-programming in specs, this can cause serious problems down the line. Write each test out

spec/lib/ruby_smb/smb1/bit_field/nt_create_andx_flags_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    expect(described_class.fields.instance_variable_get(:@hints)[:endian]).to eq :little
+  end
+
+  {
+    :request_extended_response => 0x00000010,
+    :open_target_dir           => 0x00000008,
+    :request_opbatch           => 0x00000004,
+    :request_oplock            => 0x00000002
+  }.each do |field, bitmask|
+    describe "##{field.to_s}" do
```
dmaloney-r7: again, please don't do this in rspec, I know it can be tedious to write these tests out, but meta-programming like this in rspec can cause a lot of problems

spec/lib/ruby_smb/smb1/bit_field/security_flags_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  it 'is little endian' do
+    expect(described_class.fields.instance_variable_get(:@hints)[:endian]).to eq :little
+  end
+
+  {
+    :effective_only   => 0x02,
+    :context_tracking => 0x01
+  }.each do |field, bitmask|
+    describe "##{field.to_s}" do
+      it 'is a 1-bit flag' do
```
dmaloney-r7: same here

spec/lib/ruby_smb/smb1/packet/nt_create_andx_request_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+
+    it 'has a "words" field when word_count is not zero' do
+      parameter_block.word_count = 99
+      expect(parameter_block.words?).not_to be_nil
+    end
+
+    describe '#words' do
+      subject(:words) { parameter_block.words }
+
+      it { is_expected.to respond_to :andx_command }
```
dmaloney-r7: the packet should be using the actual andx_block compound field

dmaloney-r7:
rekicked the Travis build as it seems like the build job stalled out

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
chrisdlf: Yeah, I agree with you, it is getting too much nested. Another option would be to add the `onlyif:` parameter to each line:
```ruby
and_x_block             :andx_block,         onlyif: -> { word_count.nonzero? }
# Constants defined in RubySMB::SMB1::OplockLevels
uint8                   :oplock_level,       onlyif: -> { word_count.nonzero? }, label: 'OpLock Level'
uint16                  :fid,                onlyif: -> { word_count.nonzero? }, label: 'FID'
# Constants defined in RubySMB::Dispositions
uint32                  :create_disposition, onlyif: -> { word_count.nonzero? }, label: 'Create Disposition'
file_time               :create_time,        onlyif: -> { word_count.nonzero? }, label: 'Create Time'
...
```
It is a bit of a hack, but it has the same effect without adding a layer of nesting.

I am also fine just using exception handling and the EmtpyPacket class in the client.

lib/ruby_smb/smb1/packet/nt_create_andx_request.rb
chrisdlf: Thank you for the heads-up, I really missed this one. I will fix this.

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
chrisdlf: Sounds good! (but it is still a scary protocol design :) )

lib/ruby_smb/smb1/bit_field/nt_create_andx_flags.rb
chrisdlf: Indeed, it is a good idea. Thanks for the heads-up, I will fix this.

spec/lib/ruby_smb/smb1/bit_field/file_status_flags_spec.rb
chrisdlf: Oh ok, sorry, I didn't know about this. I will fix this, no problem.
Just out of curiosity, what kind of issue this could cause?

dmaloney-r7: one of the big ones is that you won't actually be able to tell which test scenario failed if one of them fails without a lot of digging, it turns into a mess really quickly

chrisdlf: That's a very good reason, thanks!

lib/ruby_smb/smb1/packet/nt_create_andx_request.rb
chrisdlf: Addressed in 07d5066e9de72b9dce9101cf62967e1d3fde9833.

lib/ruby_smb/smb1/packet/nt_create_andx_response.rb
chrisdlf: I removed the #words and #bytes layer of nesting and also removed the `onlyif:` parameter I had originally added on each field (see 56282a3985b88f36df23c470e5f4de9f6be19c35). This can still be re-added later if you think it is worth it. Addressed in b4456bec8c6df3df4e02f7a07e14401506f2945c.

lib/ruby_smb/smb1/bit_field/nt_create_andx_flags.rb
chrisdlf: I removed this class and added an embedded struct. I also refactored the request and response specs to match the code you pointed out (https://github.com/rapid7/ruby_smb/blob/master/spec/lib/ruby_smb/smb1/packet/nt_trans/create_request_spec.rb). Addressed in d38c671976c78007294a6334ae06b8c244f22b34.

spec/lib/ruby_smb/smb1/bit_field/file_status_flags_spec.rb
chrisdlf: Addressed in 37aad9e13a382c36c0e1a23a899e6c4f549a881f.

chrisdlf:
Thanks for your feedback. I think I fixed the issues you pointed out. Please, let me know if there is anything else.

dmaloney-r7:
@chrisdlf looks good. We added SMB2 Read from file last week so you can see what we're going for 

[here](https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb2/tree.rb#L46-L95) and
[here](https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb2/file.rb)

chrisdlf:
Thanks @dmaloney-r7! I will start to work on the SMB1 Read.

PR #104, opened 2017-10-04
Add SMB1 file read, write and append functionalities

chrisdlf:
This PR adds SMB1 read, write and append functionalities.

**Client (SMB1):**
- add #max_buffer_size attribute and MAX_BUFFER_SIZE constant - the maximum size of a SMB message that the Client accepts (in bytes)
- use #max_buffer_size attribute instead of hardcoded value in lib/ruby_smb/client/authentication.rb

**Tree (SMB1)**
- add #set_header_fields helper
- update #disconnect! and #list to use #set_header_fields
- add #add_null_termination helper
- add #open_file method

**File (SMB1)**
- create the File class and adds read, write and append functionalities.

This branch also includes changes from this PR: https://github.com/rapid7/ruby_smb/pull/103.
 
# Verification Steps
- [x] `bundle install`
- [x] Verify you can read a file with `ruby examples/read_file.rb <ip-address> <username> <password> <share> <file-to-read>` (change `smb2: true` argument to `smb2: false` to force SMB1 when instantiating the Client - line 22)
- [x] Verify you can write to a file with `ruby examples/write_file.rb <ip-address> <username> <password> <share> <file-to-read> "data to write"` (change `smb2: true` argument to `smb2: false` to force SMB1 when instantiating the Client - line 23)
- [x] Verify you can append to a file with `ruby examples/append_file.rb <ip-address> <username> <password> <share> <file-to-read> "data to append"` (change `smb2: true` argument to `smb2: false` to force SMB1 when instantiating the Client - line 23)
## `rake spec`
- [x] `rake spec`
- [x] VERIFY no failures


lib/ruby_smb/client.rb
```diff
@@ -253,5 +262,39 @@ def wipe_state!
...
+    #     response_packet: RubySMB::SMB1::Packet::NtCreateAndxResponse,
+    #     raw_response: raw_response)
+    #
+    # @param response_packet [Class] the class used to parse the raw response
+    # @param raw_response [String] the raw response to parse
+    # @param expected_status_code [WindowsError::ErrorCode] the expected status code
+    # @return [RubySMB::GenericPacket] the parsed response
+    # @raise [RubySMB::Error::InvalidPacket] if the response command is not the the expected command from `response_packet`
+    # @raise [RubySMB::Error::UnexpectedStatusCode] if the response NTStatus is not `expected_status_code`
+    def parse_response(response_packet:, raw_response:, expected_status_code: WindowsError::NTStatus::STATUS_SUCCESS)
```
dmohanty-r7: All this error handling seems like it might belong in the GenericPacket class. #read does some base generic handling accounting for SMB1 and SMB2. https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/generic_packet.rb#L42 . This seems like more packet related error handling.

lib/ruby_smb/smb1/file.rb
```diff
@@ -0,0 +1,193 @@
...
+                           else
+                             bytes
+                           end
+        remaining_bytes = bytes
+        data = ''
+
+        loop do
+          read_request = read_packet(read_length: atomic_read_size, offset: offset)
+          raw_response = @tree.client.send_recv(read_request)
+          response = @tree.client.parse_response(response_packet: RubySMB::SMB1::Packet::ReadAndxResponse, raw_response: raw_response)
```
dmohanty-r7: In smb2 implementation we use the Packet class to read the response. https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb2/file.rb#L105

lib/ruby_smb/client.rb
chrisdlf: Absolutely, I removed this helper methods from the Client class and add the validation logic in the caller classes. Thanks for the heads up. Addressed in 4bb00fb.

lib/ruby_smb/smb1/file.rb
chrisdlf: I removed #parse_response helper, used the Packet class to read the response and added the validation logic. Thanks! Addressed in 4bb00fb.

PR #109, opened 2017-10-05
Add a Read Timeout to the Socket Dispatcher class

chrisdlf:
This fixes https://github.com/rapid7/ruby_smb/issues/106

# Verification Steps
- [x] `bundle install`
## `rake spec`
- [x] `rake spec`
- [x] VERIFY no failures

### DIFF
--- a/lib/ruby_smb/dispatcher/socket.rb
+++ b/lib/ruby_smb/dispatcher/socket.rb
@@ -4,15 +4,23 @@ module RubySMB
     # This class provides a wrapper around a Socket for the packet Dispatcher.
     # It allows for dependency injection of different Socket implementations.
     class Socket < RubySMB::Dispatcher::Base
+      READ_TIMEOUT = 30
+
       # The underlying socket that we select on
       # @!attribute [rw] tcp_socket
       #   @return [IO]
       attr_accessor :tcp_socket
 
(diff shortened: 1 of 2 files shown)


lib/ruby_smb/dispatcher/socket.rb
```diff
@@ -41,11 +49,15 @@ def send_packet(packet)
       # which are assumed to be the NetBiosSessionService header.
       # @return [String]
       def recv_packet
-        IO.select([@tcp_socket])
+        if IO.select([@tcp_socket], nil, nil, @read_timeout).nil?
+          raise RubySMB::Error::CommunicationError, "Read timeout expired when reading from the Socket (timeout=#{@read_timeout})"
+        end
         nbss_header = @tcp_socket.read(4) # Length of NBSS header. TODO: remove to a constant
-        raise ::RubySMB::Error::NetBiosSessionService, 'NBSS Header is missing' if nbss_header.nil?
+        raise ::RubySMB::Error::NetBiosSessionService, 'NBSS Header is missing' if nbss_header.nil? || nbss_header.empty?
```
chrisdlf: I had issues with some specific Socket implementations that return an empty string instead of nil when something wrong happens.

PR #116, opened 2017-11-09
Add SMB1 file delete and rename functionalities

chrisdlf:
This PR adds SMB1 delete and rename functionalities.

# Verification Steps
- [x] `bundle install`
- [x] Verify you can delete a file with `ruby examples/delete_file.rb <ip-address> <username> <password> <share> <file-to-delete>` (change `smb2: true` argument to `smb2: false` to force SMB1 when instantiating the Client)
- [x] Verify you can rename a file with `ruby examples/rename_file.rb <ip-address> <username> <password> <share> <old_name> <new_name>` (change `smb2: true` argument to `smb2: false` to force SMB1 when instantiating the Client)
## `rake spec`
- [x] `rake spec`
- [x] VERIFY no failures

### DIFF
--- a/examples/delete_file.rb
+++ b/examples/delete_file.rb
@@ -19,7 +19,7 @@ path     = "\\\\#{address}\\#{share}"
 sock = TCPSocket.new address, 445
 dispatcher = RubySMB::Dispatcher::Socket.new(sock)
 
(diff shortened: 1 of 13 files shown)


dmohanty-r7:
lgtm!

PR #117, opened 2017-11-30
Add attributes to client

### LINKED ISSUE #112: Add additional attributes to client
As brought to attention by @chrisdlf : Currently we are only capturing the `peer_native_os` attribute from  the SMB_COM_SESSION_SETUP_ANDX response in the client.

https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/client/authentication.rb#L37 https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/client/authentication.rb#L79

In REX we additionally get the following attributes from SMB_COM_SESSION_SETUP_ANDX
- peer_native_lm

also in REX we get these from the NTLM type 2 challenge blob:
- default_name
- dns_host_name
- default_domain
- dns_domain_name (this one has been removed recently from Rex but it was there before)

https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/proto/smb/client.rb#L901
https://github.com/rapid7/metasploit-framework/blob/master/lib/rex/proto/smb/client.rb#L561

We should be adding these additional useful attributes of information in the client.

chrisdlf:
This PR add the following attributes to the Client:
- peer_native_lm
- primary_domain
- default_name
- default_domain
- dns_host_name
- dns_domain_name
- dns_tree_name
- os_version
- dialect

Some negotiation and authentication methods/classes needed to be update in order to sets these attributes.

This fixes #112 .

# Verification Steps
- [x] `bundle install`
- [x] Verify you can authenticate to a remote server with `ruby examples/authenticate.rb <ip-address> <username> <password>`
- [x] Verify you can authenticate anonymously to a remote server with `ruby examples/anonymous_auth.rb <ip-address>`
## `rake spec`
- [x] `rake spec`
- [x] VERIFY no failures

### DIFF
--- a/lib/ruby_smb/client.rb
+++ b/lib/ruby_smb/client.rb
@@ -54,6 +54,54 @@ module RubySMB
     #   @return [String]
     attr_accessor :peer_native_os
 
+    # The Native LAN Manager of the Peer/Server.
+    # Currently only available with SMB1.
+    # @!attribute [rw] peer_native_lm
+    #   @return [String]
+    attr_accessor :peer_native_lm
+
+    # The Primary Domain of the Peer/Server.
+    # Currently only available with SMB1 and only when authentiation
+    # without NTLMSSP is used.
+    # @!attribute [rw] primary_domain
+    #   @return [String]
+    attr_accessor :primary_domain
+
+    # The Netbios Name of the Peer/Server.
+    # @!attribute [rw] default_name
+    #   @return [String]
+    attr_accessor :default_name
+
+    # The Netbios Domain of the Peer/Server.
+    # @!attribute [rw] default_domain
+    #   @return [String]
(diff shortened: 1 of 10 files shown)


lib/ruby_smb/client.rb
```diff
@@ -123,11 +171,17 @@ def initialize(dispatcher, smb1: true, smb2: true, username:, password:, domain:
       @username          = username.encode('utf-8') || ''.encode('utf-8')
       @max_buffer_size   = MAX_BUFFER_SIZE
 
+      negotiate_version_flag = 0x02000000
```
chrisdlf: The NEGOTIATE_VERSION flag is missing in rubyntlm library. I had to manually add it. I also submitted a PR to add it: https://github.com/WinRb/rubyntlm/pull/38

PR #118, opened 2017-12-06
NetShareEnumAll over DCE/RPC

### LINKED ISSUE #100: Implement DCERPC command to enumerate share
We need to build the basics infrastructure for DCERPC over Named Pipes on the IPC$ share. The first DCERPC method we should implement is NetShareEnum to allow the client to enumerate what shares are available before making a TreeConnect

dmohanty-r7:
This fixes #100

This adds DCE/RPC structs needed to call SRVSVC's NetShareEnumAll over SMB2 IOCTL requests.

## Validation
- [x] Target an SMB2 enabled machine
- [x] `ruby examples/net_share_enum_all.rb <ipaddress> "User Name" password`

## TODO
* Allow for IOCTL requests to be substituted with NT_TRANSACT for SMB1 support
* Implement NDR structs as opposed to manually unpacking the response
* Consider moving share enumeration functionality out of the client/library (but keep it in an example script to demonstrate dcerpc)
* Fill in missing Comments / Tests


lib/ruby_smb/dcerpc/srvsvc.rb
```diff
@@ -0,0 +1,96 @@
...
+        endian :little
+
+        uint32 :referent_id, initial_value: 0x00000001
+        uint32 :max_count,    initial_value: -> { (server_unc.do_num_bytes / 2) + 1 }
+        uint32 :offset,       initial_value: 0
+        uint32 :actual_count, initial_value: -> {max_count}
+        string :server_unc,   pad_front: false,
+               initial_value: -> {host.encode('utf-16le')}
+
+        uint16 :padding, initial_value: 0
```
chrisdlf: I think we should calculate the proper padding to follow the NDR alignement specifications:
- [Alignment of Primitive Types](http://pubs.opengroup.org/onlinepubs/9629399/chap14.htm#tagcjh_19_02_02)
- [Alignment of Constructed Types](http://pubs.opengroup.org/onlinepubs/9629399/chap14.htm#tagcjh_19_03_02)

lib/ruby_smb/dcerpc/srvsvc.rb
```diff
@@ -0,0 +1,96 @@
...
+        uint32 :ctr_referent_id, initial_value: 0x00000001
+        uint32 :ctr_count, initial_value: 0
+        uint32 :pointer_to_array, initial_value: 0
+
+        uint32 :max_buffer, initial_value: 4294967295
+
+        uint32 :resume_referent_id, value: 0x00000001
+        uint32 :resume_handle, initial_value: 0
+
+        def self.parse_response(response)
```
chrisdlf: I am wondering if it is possible to handle this with only BinData structures and avoid slicing and unpacking the directly the byte stream?

lib/ruby_smb/smb2/packet/ioctl_response.rb
```diff
@@ -17,7 +17,7 @@ class IoctlResponse < RubySMB::GenericPacket
         uint32        :output_count,        label: 'Output Count'
         uint32        :flags,               label: 'Flags'
         uint32        :reserved2,           label: 'Reserved Space'
-        string        :buffer,              label: 'Input Buffer'
+        string        :buffer,              label: 'Input Buffer',        length: -> { output_count }
```
chrisdlf: You might want to add the `:read_length` parameter to avoid issue when reading:
```ruby
       string        :buffer,              label: 'Input Buffer', read_length: -> { input_count + output_count }
```

lib/ruby_smb/smb2/packet/ioctl_response.rb
```diff
@@ -17,7 +17,7 @@ class IoctlResponse < RubySMB::GenericPacket
...
```
chrisdlf: Also, it might be a good idea to add helper methods to extract input and output data (see [2.2.32 SMB2 IOCTL Response](https://msdn.microsoft.com/en-us/library/cc246548.aspx)):
```ruby
        def input_data
          to_binary_s[input_offset, input_count]
        end

        def output_data
          to_binary_s[output_offset, output_count]
        end
```

lib/ruby_smb/dcerpc/handle.rb
```diff
@@ -0,0 +1,45 @@
...
+        request.ctl_code = 0x0011C017
+        request.flags.is_fsctl = 0x00000001
+        request.buffer = action.to_binary_s
+        @last_msg = @pipe.tree.client.send_recv(request)
+        handle_msg(RubySMB::SMB2::Packet::IoctlResponse.read(@last_msg))
+      end
+
+      def handle_msg(msg)
+        if @msg_type == :request
+          dcerpc_response_stub = RubySMB::Dcerpc::Response.read(msg.buffer.to_binary_s).stub
```
chrisdlf: In my understanding, the output data of a IOCTL Response could be at a different offset in `buffer` ([2.2.32 SMB2 IOCTL Response](https://msdn.microsoft.com/en-us/library/cc246548.aspx)). The `OutputOffset` and `OutputCount` fields can be used to extract the correct output data.
Please have a look at the helper methods I suggest in my comment below (for the lib/ruby_smb/smb2/packet/ioctl_response.rb file)

lib/ruby_smb/dcerpc/srvsvc.rb
```diff
@@ -0,0 +1,96 @@
...
+
+        uint32 :ctr, initial_value: 1
+        uint32 :ctr_referent_id, initial_value: 0x00000001
+        uint32 :ctr_count, initial_value: 0
+        uint32 :pointer_to_array, initial_value: 0
+
+        uint32 :max_buffer, initial_value: 4294967295
+
+        uint32 :resume_referent_id, value: 0x00000001
+        uint32 :resume_handle, initial_value: 0
```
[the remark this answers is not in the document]
dmohanty-r7: This is a great start for an NDR implementation. I played around with this for the server unc, but I think I will make an issue for this to address it separately from this PR.

lib/ruby_smb/dcerpc/srvsvc.rb
dmohanty-r7: same as above^

PR #120, opened 2018-01-02
Add client methods

dmohanty-r7:
This is a WIP which adds various general methods and attributes to the smb client that metasploit-framework uses.

It adds convenience methods for opening/reading/writing to files from the client. It also adds attributes to the client. (there are also commits from the DCERPC work #118)


busterb:
Looks good to me, thanks!

PR #125, opened 2018-02-09
Raise an exception if the socket closes unexpectedly

asoto-r7:
The `ms17_010_eternalblue` module was occasionally causing a `nil` to be returned from the `tcp_socket.write()`.  This PR adds exception handling and reports a `RubySMB::Error::CommunicationError` so it can be handled by the module.

In a related PR, the Metasploit project will need to modify `ms17_010_eternalblue` module handles this exception cleanly.  See comments for that matching PR.

### DIFF
--- a/lib/ruby_smb/dispatcher/socket.rb
+++ b/lib/ruby_smb/dispatcher/socket.rb
@@ -40,6 +40,13 @@ module RubySMB
           while bytes_written < data.size
             bytes_written += @tcp_socket.write(data[bytes_written..-1])
           end
+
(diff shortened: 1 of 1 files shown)


asoto-r7:
Hmmm, I seem to have angered the tests.  Is this a real problem, or do we need to fix the test cases?

screen shot 2018-02-09 at 2 56 15 pm

asoto-r7:
I'm dumb.  I missed a line, which caused the test to fail.  But now it's good.

PR #126, opened 2018-02-22
store negotiated max server buffer size

bcook-r7:
This is something that is used by the new named pipe functionality, so let's pull it in at the client level.

### DIFF
--- a/lib/ruby_smb/client/negotiation.rb
+++ b/lib/ruby_smb/client/negotiation.rb
@@ -80,24 +80,19 @@ module RubySMB
         when RubySMB::SMB1::Packet::NegotiateResponseExtended
           self.smb1 = true
           self.smb2 = false
-          if packet.parameter_block.security_mode.security_signatures_required == 1
-            self.signing_required = true
-          else
-            self.signing_required = false
-          end
+          self.signing_required = packet.parameter_block.security_mode.security_signatures_required == 1
           self.dialect = packet.negotiated_dialect.to_s
+          self.server_max_buffer_size = packet.parameter_block.max_buffer_size
           'SMB1'
         when RubySMB::SMB2::Packet::NegotiateResponse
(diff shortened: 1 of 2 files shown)


lib/ruby_smb/client/negotiation.rb
```diff
@@ -80,24 +80,19 @@ def parse_negotiate_response(packet)
...
           self.smb1 = false
           self.smb2 = true
-          self.signing_required = if packet.security_mode.signing_required == 1
-                                    true
-                                  else
-                                    false
-                                  end
+          self.signing_required = packet.security_mode.signing_required == 1
           self.dialect = "0x%04x" % packet.dialect_revision
+          self.server_max_buffer_size = 16644
```
bcook-r7: I wasn't sure where to get this with SMB2, so I just made it the default value that a server with > 512MB of RAM would have, which seemed like enough for anyone (tm)

UserExistsError: https://github.com/rapid7/ruby_smb/blob/607966396d5a0247ecbfc43dc635d2a4cc76aece/lib/ruby_smb/smb2/packet/negotiate_response.rb#L17
For SMB2 pipes we'll need max_write_size and max_read_size. It appears SMB1 MaxBufferSize was split into max_write_size, max_read_size, and max_transaction_size for SMB2.

busterb: ah, thanks. I'll put out an updated PR.

busterb: Well, that is unless you want to @UserExistsError - the main goal here is to get ruby_smb integrated in simple client in MSF, which needs some updating to work with bind_named_pipe additions

UserExistsError: i'll take a look

PR #127, opened 2018-02-25
simpleclient, namedpipe progress

UserExistsError:
This PR contains progress towards replacing Rex in simpleclient. Specifically, it adds pipe functionality for the bind_named_pipe payload.

## Changes
1) Added missing "reserved" field to SMB2 TREE_DISCONNECT Requests
2) Added SMB1::Pipe and SMB2::Pipe which subclass File and provide pipe specific operations such as "peek". Correct instance of File or Pipe is returned by Tree.open_file.
3) Nbss::SessionHeader length field increased from 16 to 17 bits. For SMB2 this field can be 24 bits. This change allows server max read/write/transact sizes to be used as is (they are typically 64KiB).
   References
      SMB1 https://msdn.microsoft.com/en-us/library/cc246249.aspx
      SMB2 https://msdn.microsoft.com/en-us/library/cc246496.aspx
4) Save off max read/write/transact sizes from SMB2 negotiation. It isn't clear to me how/if these apply differently to files vs pipes. For files, larger requests can be issued (observed with smbclient). Enforcing this limit for both, for now at least.
5) Added examples/pipes.rb to demonstrate new functionality.

## Future Improvements
1) Async read/writes. This should improve bind_named_pipe comms by not having to constantly poll metsrv (get rid of peeking). The current model is to send a request then wait for the response, as it is in Rex. This is an issue for the bind_named_pipe payload because the packet dispatcher has multiple threads trying to read and write. Pipe reads/writes must be synchronized to get bind_named_pipe comms to work. I still need a better understanding of this issue.

## Refs
https://github.com/rapid7/metasploit-framework/pull/9365

### DIFF
--- a/lib/ruby_smb/smb1/packet/trans.rb
+++ b/lib/ruby_smb/smb1/packet/trans.rb
@@ -8,8 +8,8 @@ module RubySMB
         require 'ruby_smb/smb1/packet/trans/data_block'
         require 'ruby_smb/smb1/packet/trans/request'
         require 'ruby_smb/smb1/packet/trans/response'
-        require 'ruby_smb/smb1/packet/trans/peek_named_pipe_request'
-        require 'ruby_smb/smb1/packet/trans/peek_named_pipe_response'
+        require 'ruby_smb/smb1/packet/trans/peek_nmpipe_request'
+        require 'ruby_smb/smb1/packet/trans/peek_nmpipe_response'
       end
     end
   end
--- a/lib/ruby_smb/smb1/packet/trans.rb
+++ b/lib/ruby_smb/smb1/packet/trans.rb
@@ -1,0 +1,16 @@
+module RubySMB
+  module SMB1
+    module Packet
+      # Namespace for the Transaction sub-protocol documented in
+      # [2.2.4.33 SMB_COM_TRANSACTION (0x25)](https://msdn.microsoft.com/en-us/library/ee441489.aspx)
+      module Trans
+        require 'ruby_smb/smb1/packet/trans/subcommands'
+        require 'ruby_smb/smb1/packet/trans/data_block'
+        require 'ruby_smb/smb1/packet/trans/request'
+        require 'ruby_smb/smb1/packet/trans/response'
+        require 'ruby_smb/smb1/packet/trans/peek_named_pipe_request'
+        require 'ruby_smb/smb1/packet/trans/peek_named_pipe_response'
+      end
+    end
+  end
+end
--- a/lib/ruby_smb/smb2/pipe.rb
+++ b/lib/ruby_smb/smb2/pipe.rb
@@ -20,18 +20,10 @@ module RubySMB
         packet.max_output_response = 16 + peek_size
         packet = set_header_fields(packet)
         raw_response = @tree.client.send_recv(packet)
-        begin
-          response = RubySMB::SMB2::Packet::IoctlResponse.read(raw_response)
-        rescue EOFError
-          raise RubySMB::Error::InvalidPacket, 'Failed to process IoctlResponse packet'
-        end
+        response = RubySMB::SMB2::Packet::IoctlResponse.read(raw_response)
 
         unless response.smb2_header.command == RubySMB::SMB2::Commands::IOCTL
           raise RubySMB::Error::InvalidPacket, 'Not an IoctlResponse packet'
-        end
-
-        unless response.ctl_code == RubySMB::Fscc::ControlCodes::FSCTL_PIPE_PEEK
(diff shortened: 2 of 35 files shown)


dmohanty-r7:
is lib/ruby_smb/smb1/packet/trans/peek_named_pipe.rb needed?

lib/ruby_smb/smb2/pipe.rb
```diff
@@ -0,0 +1,39 @@
...
+      # @return [RubySMB::SMB2::Packet::IoctlResponse]
+      def peek(peek_size: 0)
+        packet = RubySMB::SMB2::Packet::IoctlRequest.new
+        packet.ctl_code = RubySMB::Fscc::ControlCodes::FSCTL_PIPE_PEEK
+        packet.flags.is_fsctl = true
+        # read at least 16 bytes for state, avail, msg_count, first_msg_len
+        packet.max_output_response = 16 + peek_size
+        packet = set_header_fields(packet)
+        resp = @tree.client.send_recv(packet)
+        RubySMB::SMB2::Packet::IoctlResponse.read(resp)
```
dmohanty-r7: Sometimes we will get Error packets (for smb2) or Empty packets (for smb2), not being what is expected (RubySMB::SMB2::Packet::IoctlResponse) we might want to handle the cases here or raise an exception.

UserExistsError:
Removed lib/ruby_smb/smb1/packet/trans/peek_named_pipe.rb and added error handling to SMB1::Pipe and SMB2::Pipe

lib/ruby_smb/smb2/pipe.rb
```diff
@@ -11,29 +11,45 @@ class Pipe < File
...
         packet.flags.is_fsctl = true
         # read at least 16 bytes for state, avail, msg_count, first_msg_len
         packet.max_output_response = 16 + peek_size
         packet = set_header_fields(packet)
-        resp = @tree.client.send_recv(packet)
-        RubySMB::SMB2::Packet::IoctlResponse.read(resp)
+        raw_response = @tree.client.send_recv(packet)
+        begin
+          response = RubySMB::SMB2::Packet::IoctlResponse.read(raw_response)
+        rescue EOFError
```
dmohanty-r7: A begin/rescue EOFError like this happens in https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/generic_packet.rb#L45 returning either a RubySMB::SMB1::Packet::EmptyPacket or RubySMB::SMB1::Packet::ErrorPacket so you don't need to do that here.

lib/ruby_smb/smb2/pipe.rb
```diff
@@ -0,0 +1,55 @@
...
+      # @raise [RubySMB::Error::InvalidPacket] if not a valid FSCTL_PIPE_PEEK response
+      def peek(peek_size: 0)
+        packet = RubySMB::SMB2::Packet::IoctlRequest.new
+        packet.ctl_code = RubySMB::Fscc::ControlCodes::FSCTL_PIPE_PEEK
+        packet.flags.is_fsctl = true
+        # read at least 16 bytes for state, avail, msg_count, first_msg_len
+        packet.max_output_response = 16 + peek_size
+        packet = set_header_fields(packet)
+        raw_response = @tree.client.send_recv(packet)
+        begin
```
dmohanty-r7: this begin/rescue is not needed since #read will attempt to at least get you the header via an ErrorPacket https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/generic_packet.rb#L45

lib/ruby_smb/smb1/packet/trans.rb
```diff
@@ -0,0 +1,16 @@
+module RubySMB
```
dmohanty-r7: @chrisdlf has a branch right now that currently also implements trans/trans2 packets in a very similar way to this, also including spec tests with a few minor differences. We like your implementation and wanted to see what your thoughts were: https://github.com/UserExistsError/ruby_smb/compare/simpleclient-progress...chrisdlf:feature/100/dcerpc_smb1_named_pipe?expand=1#diff-42e45c6d8a4c2b5cd3c864fcccc34337 . We can modify his branch to work with this, we just wanted to ask if you had any thoughts about the different approaches.

UserExistsError: That implementation looks much more complete. I renamed a few things in my last commit to follow naming conventions used in that branch.

dmohanty-r7:
@UserExistsError The code looks good to me. One thing though: recently we have dropped below the ideal % test coverage, but we would like to keep it close to 95%. Some spec tests would be nice to have for new functionality.

dmohanty-r7:
Looks good, thank you @UserExistsError! I will add an issue for the asynchronous reading and writing.

PR #130, opened 2018-03-08
Fix SMB2 DCERPC bind request with Windows 10

chrisdlf:
This PR includes:

**1. Use SMB2 Write/Read instead of IOCTL request to send the DCERPC Bind request**
After multiple tests with different Windows versions, I noticed that the Bind request use a standard SMB2 Write/Read function to send the Bind request and retrieve the BindAck response.

**2. Change the default value of `max_output_response` for `IoctlRequest`**
This was causing issue with recent Windows (STATUS_INVALID_PARAMETER).


dmohanty-r7:
Nice find @chrisdlf

PR #134, opened 2018-04-27
Error fixes

jrobles-r7:
Catch IOError that occurs when parsing an ErrorPacket as ResponsePacket.
Remove unused access variable in utils open.
Add STATUS_PENDING checks.

### DIFF
--- a/lib/ruby_smb/generic_packet.rb
+++ b/lib/ruby_smb/generic_packet.rb
@@ -42,7 +42,7 @@ module RubySMB
     def self.read(val)
       begin
         super(val)
-      rescue EOFError, IOError => e
+      rescue IOError => e
         case self.to_s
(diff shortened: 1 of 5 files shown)


lib/ruby_smb/generic_packet.rb
```diff
@@ -42,7 +42,7 @@ def packet_smb_version
     def self.read(val)
       begin
         super(val)
-      rescue EOFError => e
+      rescue EOFError, IOError => e
```
chrisdlf: If I remember well, EOFError is a descendant of the IOError class. So, I think using `rescue IOError => e` should be enough.

chrisdlf:
Looks good to me. I think this PR fixes is related to this issue: https://github.com/rapid7/ruby_smb/issues/128.

PR #138, opened 2018-07-03
don't pass a socket straight into bindata read

busterb:
Seekability on a _socket_ is OS-dependent, so we should instead read what we want into a string/buffer and pass that in instead. This was the only occurence I could find. Fixes support on Windows. Without this, we get the following backtrace on the first send_recv call on a Windows host:

```
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/io.rb:31:in `pos'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/io.rb:31:in `seekable?'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/io.rb:23:in `initialize'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/io.rb:240:in `initialize'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/base.rb:143:in `new'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/base.rb:143:in `read'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/bindata-2.4.3/lib/bindata/base.rb:21:in `read'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/ruby_smb-1.0.2/lib/ruby_smb/dispatcher/socket.rb:69:in `recv_packet'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/ruby_smb-1.0.2/lib/ruby_smb/client.rb:334:in `send_recv'
c:/metasploit-framework/embedded/lib/ruby/gems/2.3.0/gems/ruby_smb-1.0.2/lib/ruby_smb/client/negotiation.rb:14:in `negotiate'
c:/metasploit-framework/embedded/framework/lib/rex/proto/smb/simpleclient.rb:77:in `login'
c:/metasploit-framework/embedded/framework/lib/msf/core/exploit/smb/client.rb:132:in `smb_login'
c:/metasploit-framework/embedded/framework/modules/exploits/windows/smb/psexec.rb:94:in `exploit'
```

### DIFF
--- a/lib/ruby_smb/dispatcher/socket.rb
+++ b/lib/ruby_smb/dispatcher/socket.rb
@@ -66,7 +66,7 @@ module RubySMB
         end
 
         begin
-          nbss_header = RubySMB::Nbss::SessionHeader.read(@tcp_socket)
+          nbss_header = RubySMB::Nbss::SessionHeader.read(@tcp_socket.read(4))
         rescue IOError
           raise ::RubySMB::Error::NetBiosSessionService, 'NBSS Header is missing'
         end
--- a/lib/ruby_smb/dispatcher/socket.rb
+++ b/lib/ruby_smb/dispatcher/socket.rb
@@ -66,7 +66,9 @@ module RubySMB
         end
 
         begin
-          nbss_header = RubySMB::Nbss::SessionHeader.read(@tcp_socket.read(4))
+          nbss_data = @tcp_socket.read(4)
+          raise IOError if nbss_data.nil?
+          nbss_header = RubySMB::Nbss::SessionHeader.read(nbss_data)
         rescue IOError
           raise ::RubySMB::Error::NetBiosSessionService, 'NBSS Header is missing'
         end
--- a/spec/lib/ruby_smb/dispatcher/socket_spec.rb
+++ b/spec/lib/ruby_smb/dispatcher/socket_spec.rb
@@ -85,7 +85,7 @@ RSpec.describe RubySMB::Dispatcher::Socket do
       end
 
       it 'reads the nbss header using Nbss::SessionHeader structure' do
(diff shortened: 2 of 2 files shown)


busterb:
Does this look ok to you @chrisdlf ?

busterb:
Hmm, this needs a little more love, will update once I've got it working.

chrisdlf:
Hi @busterb,
I did some investigation and I found that BinData already checks if the IO stream is "seekable" by trying to call _#pos_ method (https://github.com/dmendel/bindata/blob/master/lib/bindata/io.rb#L23 and https://github.com/dmendel/bindata/blob/master/lib/bindata/io.rb#L31). But, apparently, the exception thrown is different on Windows (Errno::EINVAL) and the library is not catching it: https://github.com/dmendel/bindata/blob/master/lib/bindata/io.rb#L32

So, I did some test and patch BinData to make it work:
```ruby
      def seekable?
        @raw_io.pos
      rescue NoMethodError, Errno::ESPIPE, Errno::EPIPE, Errno::EINVAL
        nil
      end
```

Maybe we could submit this fix instead of patching RubySMB?

busterb:
Great observation @chrisdlf. Maybe we can do both for now. I mainly suggest this because apparently this is breaking Metasploit's ability to use the latest version of ruby_smb on windows, which we need to fix this week.

chrisdlf:
Awesome! Thanks for submitting this PR.

PR #143, opened 2018-10-11
Update EmptyPacket and ErrorPacket Validation

chrisdlf:
This PR adds some specific checks in the `valid?` method for `EmptyPacket` and `ErrorPacket`. It also updates the fields of SMB2 `ErrorPacket` to follow the protocol . This will ensure the validation to fail if the error packet is malformed.

Before these changes, when the response was not the expected packet but the protocol ID and the command ID were correct, it was being converted to a **valid** `EmptyPacket` or `ErrorPacket`, even if it was not an error packet at all.

I came across this corner case when I was probing an Echo service. The SMB request was returned verbatim as a response, which, as a result, included the correct protocol ID and command ID. Since it was not the expected packet structure, the parsing was failing and the response was converted to an error packet, but `valid?` was returning true (protocol ID and command ID were correct). Now, `valid?` will return false in this situation, since the extra checks will fail.

### DIFF
--- a/lib/ruby_smb/smb1/packet/empty_packet.rb
+++ b/lib/ruby_smb/smb1/packet/empty_packet.rb
@@ -1,7 +1,7 @@
 module RubySMB
   module SMB1
     module Packet
-      # This packet represent an SMB1 Response Packet when the parameter and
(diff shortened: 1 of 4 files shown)


jrobles-r7:
Thanks Christophe!

PR #147, opened 2019-03-12
style consistency fixes, some grammar and spelling

busterb:
This also changes all of the example code to use the defaults for smb1/smb2 rather than disabling smb2 everywhere, which seems to better match the explanations about how the user does not have to care what version to support since it is automatic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,12 @@
 [![Code Climate](https://codeclimate.com/github/rapid7/ruby_smb.png)](https://codeclimate.com/github/rapid7/ruby_smb)
 [![Coverage Status](https://coveralls.io/repos/github/rapid7/ruby_smb/badge.svg?branch=master)](https://coveralls.io/github/rapid7/ruby_smb?branch=master)
 
-A native Ruby implementation of the SMB Protocol Family. It currently supports
+This is a native Ruby implementation of the SMB Protocol Family. It currently supports:
+
  1. [[MS-SMB]](https://msdn.microsoft.com/en-us/library/cc246231.aspx)
  1. [[MS-SMB2]](http://msdn.microsoft.com/en-us/library/cc246482.aspx)
 
-This library currently include both a client level, and packet level support. A user can parse and manipulate raw SMB packets, or use the simple client to perform SMB operations.
+The RubySMB library provides client-level and packet-level support for the protocol. A user can parse and manipulate raw SMB packets, or use the client to perform higher-level SMB operations.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,11 +4,12 @@
 [![Code Climate](https://codeclimate.com/github/rapid7/ruby_smb.png)](https://codeclimate.com/github/rapid7/ruby_smb)
 [![Coverage Status](https://coveralls.io/repos/github/rapid7/ruby_smb/badge.svg?branch=master)](https://coveralls.io/github/rapid7/ruby_smb?branch=master)
 
-A native Ruby implementation of the SMB Protocol Family. It currently supports
+This is a native Ruby implementation of the SMB Protocol Family. It currently supports
```
bcoles:
```suggestion
This is a native Ruby implementation of the SMB Protocol Family. It currently supports:
```

Also, `Family` is capitalized?

README.md
```diff
@@ -4,11 +4,12 @@
...
+
  1. [[MS-SMB]](https://msdn.microsoft.com/en-us/library/cc246231.aspx)
  1. [[MS-SMB2]](http://msdn.microsoft.com/en-us/library/cc246482.aspx)
 
-This library currently include both a client level, and packet level support. A user can parse and manipulate raw SMB packets, or use the simple client to perform SMB operations.
+The RubySMB library provides client level and packet level support for the protocol. A user can parse and manipulate raw SMB packets, or use the client to perform higher-level SMB operations.
```
bcoles: Should `client-level` and `packet-level` be hyphenated?

```suggestion
The RubySMB library provides client-level and packet-level support for the protocol. A user can parse and manipulate raw SMB packets, or use the client to perform higher-level SMB operations.
```

README.md
```diff
@@ -190,38 +180,35 @@ Example:
...
   client.negotiate
 ```
 
 #### Authentication
 
-Authentication is achieved via the [ruby ntlm gem](https://rubygems.org/gems/rubyntlm). While the client
-will not currently attempt older basic authentication on its own, it will attempt an anonymous login, if no
+RubySMB uses the [Ruby NTLM gem](https://rubygems.org/gems/rubyntlm) for authentication. While the client
+will not currently attempt older basic authentication on its own, it will attempt an anonymous login if no
 user credentials are supplied:
```
bcoles:
```suggestion
user credentials are supplied.
```

busterb:
Thanks for the suggestions, I merged and squashed.

PR #150, opened 2019-05-22
Fix client pipe example script

jrobles-r7:
While testing the Winreg PR I discovered the `examples/pipes.rb` was failing due to previous changes.

- Revert return value from `#open`. It had [previously changed](https://github.com/rapid7/ruby_smb/commit/9a391ea44781ed803924c7e43629e8b7aecefa65#diff-f36d39255face0869afcae76098ec44dR35) for msf compatibility (so a solution will be needed for this on the msf side)
- Update `pipes.rb` to remove `nil` argument since a [previous commit](https://github.com/rapid7/ruby_smb/commit/dd463fcda4ee7ffc199f382fb414e826b4c47ec2#diff-f36d39255face0869afcae76098ec44dL38) removed an unused parameter from `#create_pipe`

### Testing

`ruby examples/pipes.rb <ip> '\netlogon' <username> <password> <smbversion>`

### DIFF
--- a/examples/pipes.rb
+++ b/examples/pipes.rb
@@ -33,7 +33,8 @@ end
 
 client.authenticate
 client.tree_connect("\\\\#{address}\\IPC$")
-pipe = client.create_pipe(pipename)
+client.create_pipe(pipename)
+pipe = client.last_file
 
 puts "Available:       #{pipe.peek_available}"
 puts "PipeState:       #{pipe.peek_state}" # 3 == OK
--- a/examples/pipes.rb
+++ b/examples/pipes.rb
@@ -33,7 +33,7 @@ end
 
 client.authenticate
 client.tree_connect("\\\\#{address}\\IPC$")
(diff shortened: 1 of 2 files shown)


examples/pipes.rb
```diff
@@ -33,7 +33,7 @@
 
 client.authenticate
 client.tree_connect("\\\\#{address}\\IPC$")
-pipe = client.create_pipe(pipename, nil)
+pipe = client.create_pipe(pipename)
```
chrisdlf: Since #create_pipe is only used in this example, I think we could just update it instead of modifying the actual method return value, which would require some changes on msf side. Maybe something like this would be enough:
```ruby
client.create_pipe(pipename)
pipe =  client.last_file
```

jrobles-r7: I updated the return value as you mentioned.

jrobles-r7:
Since this is just an example change now I'll land it.

PR #151, opened 2019-05-29
Bump version to 1.1.0

jrobles-r7:


jrobles-r7:
@bcook-r7 is this the correct way to bump the minor version before running the jenkins job?

PR #155, opened 2020-06-10
Store server time values and GUID

zeroSteiner:
This stores the server start and system times in the client object. This will be necessary for upcoming changes to Metasploit's `smb_version` module which will consolidate the `auxilary/scanner/smb/smb2` module. The `smb2` scanner module currently reports the amount of time the remote system has been booted, so to not lose this we'll need it in RubySMB.

For testing you can inspect the object manually, or check out the [`smb_version` WIP](https://github.com/zeroSteiner/metasploit-framework/blob/36ba47398a263de0943fef5a0312b7dd3628ffe7/modules/auxiliary/scanner/smb/smb_version.rb). If you choose to go with the later, `uptime` should be reported for any server that reports *both* the system and start times.

### DIFF
--- a/lib/ruby_smb/client/negotiation.rb
+++ b/lib/ruby_smb/client/negotiation.rb
@@ -25,7 +25,7 @@ module RubySMB
           parse_smb3_encryption_data(request_packet, response_packet)
         end
 
-        if %w{0x0202 0x0210 0x0300 0x0302 0x0311 0x02ff}.include? @dialect
+        if response_packet && %w{0x0202 0x0210 0x0300 0x0302 0x0311 0x02ff}.include?(@dialect)
           if response_packet.server_start_time != 0
             @server_start_time = response_packet.server_start_time.to_time
           end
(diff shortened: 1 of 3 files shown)


lib/ruby_smb/client/negotiation.rb
```diff
@@ -24,6 +24,17 @@ def negotiate
...
+
+        if response_packet && %w{0x0202 0x0210 0x0300 0x0302 0x0311 0x02ff}.include?(@dialect)
+          @server_guid = response_packet.server_guid
+          if response_packet.server_start_time != 0
+            @server_start_time = response_packet.server_start_time.to_time
+          end
+          if response_packet.system_time != 0
+            @server_system_time = response_packet.system_time.to_time
+          end
+        end
```
cdelafuente-r7, line 36: Would it be possible to add this logic to the `parse_negotiate_response` method?
I believe this can be added under the `when RubySMB::SMB2::Packet::NegotiateResponse` case directly, which indirectly check the SMB version (no need to check the dialect in this case).

zeroSteiner:
Moved to `parse_negotiate_response` in commit 2a61d9e. Thanks for the review!

cdelafuente-r7:
I tested with the smb_version WIP module against multiple Windows versions and verified that `uptime` and `server_guid` information was reported each time.

PR #157, opened 2020-06-22
Fix Session Signing For SMB 3.1.1

zeroSteiner:
I noticed a bug that was causing the `SMB::AlwaysEncrypt` setting in Metasploit to not be honored. It turns out that with the dialect is SMB 3.1.1 and the encryption algorithms are passed as part of the negotiation capabilities, the flag is not set. This is at least the behavior I observed while testing Windows 10 1809 x64. Per [MS-SMB2](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/63abf97c-0d09-47e2-88d6-6bfa552949a5) the `SMB2_GLOBAL_CAP_ENCRYPTION` is only valid for the `3.0` and `3.0.2` dialects. This caused session encryption to always be disabled for 3.1.1 servers, there by ignoring the `SMB::AlwaysEncrypt` setting.

My proposed solution for this is to move the `@session_encrypt_data` assignment until after the SMB3 capabilities have been set (and by extension the `@encryption_algorithm` attribute). We can then check if this attribute is not nil to identify that the server supports encryption and we can enable session encryption if it was requested by the user.

While investigating this issue I also noticed that the `parse_smb3_encryption_data` function was parsing and processing the encryption, compression and preauth integrity capabilities so I renamed it to more accurately represent what it does.

## Testing

For testing:

- [ ] Verify a session can be opened to a 3.1.1 server and that it uses session encryption
- [ ] Verify a session can be opened to either a 3.0 or 3.0.2 server and that it uses session encryption

### DIFF
--- a/lib/ruby_smb/client/negotiation.rb
+++ b/lib/ruby_smb/client/negotiation.rb
@@ -18,15 +18,10 @@ module RubySMB
         # This is only valid for SMB1.
         response_packet.dialects = request_packet.dialects if response_packet.respond_to? :dialects=
         version = parse_negotiate_response(response_packet)
-        case @dialect
-        when '0x0300', '0x0302'
-          if response_packet&.capabilities&.encryption == 1
-            @encryption_algorithm = RubySMB::SMB2::EncryptionCapabilities::ENCRYPTION_ALGORITHM_MAP[RubySMB::SMB2::EncryptionCapabilities::AES_128_CCM]
-          end
-        when '0x0311'
-          parse_smb3_capabilities(request_packet, response_packet)
+        if @dialect == '0x0311'
+          update_preauth_hash(request_packet)
+          update_preauth_hash(response_packet)
         end
-        @session_encrypt_data = self.smb3 && @session_encrypt_data && !@encryption_algorithm.nil?
 
         # If the response contains the SMB2 wildcard revision number dialect;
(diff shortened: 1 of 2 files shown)


zeroSteiner:
image

Yeah so the change I added broke the unit tests, so I did quite a bit of refactoring work in 38a0c1f to fix them up and get them back running smoothly.

One major thing I did was move the attribute assignments into `parse_negotiate_response`. This mean that `parse_smb3_capabilities` had to be moved into there and because the `request` wasn't available, that was moved higher up into `negotiate` which has access to both the request and response. This then meant I had a bunch of reworking to do in the specs. I hope this looks good, I wouldn't be opposed to an alternative approach.

cdelafuente-r7:
Oh! I added [this line](https://github.com/rapid7/ruby_smb/commit/8c381b148b6119ada407a30421a75e422ea1e7f7#diff-437d0227c3c5b43448d4a70b7947caf3R133) as an improvement and turned out I introduced a bug by misreading the specs. Thanks for catching this!

zeroSteiner:
Thanks for the review @cdelafuente-r7! I updated the specs per your suggestions.

cdelafuente-r7:
Thanks @zeroSteiner! It is merged now.

PR #164, opened 2020-09-17
Minimal Netlogon Implementation For CVE-2020-1472 (AKA: Zerologon)

zeroSteiner:
This implements the basic functionality for exploiting CVE-2020-1472, also known as Zerologon which is a vulnerability within the authentication process of the Netlogon service.

This adds two request and response pairs as well as a function for calculating the session key used for encryption.

I'll be opening a PR to Metasploit for the module which will leverage this later today. The easiest way to test this would be to use that module which takes advantage of all of the functionality that I add here.

### DIFF
--- a/lib/ruby_smb/dcerpc/ndr.rb
+++ b/lib/ruby_smb/dcerpc/ndr.rb
@@ -95,6 +95,85 @@ module RubySMB
         end
       end
 
+      # An NDR Uni-dimensional Fixed Array of bytes representation as defined in:
+      # [Transfer Syntax NDR - NDR Constructed Types](https://pubs.opengroup.org/onlinepubs/9629399/chap14.htm#tagcjh_19_03_03_01)
+      class NdrFixedByteArray < BinData::BasePrimitive
+        optional_parameters :length, :pad_byte, :pad_front
+        default_parameters pad_byte: 0
+        mutually_exclusive_parameters :length, :value
+
+        def assign(val)
+          super(fixed_byte_array(val))
+        end
+
+        def snapshot
+          clamp_to_length(super)
+        end
+
+        class << self
+          def arg_processor
+            NdrFixedByteArrayArgProcessor.new
+          end
+        end
+
+        private
+
+        def clamp_to_length(val)
+          val = fixed_byte_array(val)
+          len = eval_parameter(:length) || val.length
+          if val.length > len
+            val = val.first(len)
+          elsif val.length < len
+            pad = eval_parameter(:pad_byte)
+            if get_parameter(:pad_front)
+              val = val.insert(0, *Array.new(len - val.length, pad))
+            else
+              val = val.fill(pad, val.length...len)
+            end
+          end
+
+          val
+        end
+
+        def fixed_byte_array(val)
+          val = val.bytes if val.is_a? String
+          val.to_ary
+        end
+
+        def read_and_return_value(io)
+          len = eval_parameter(:length) || 0
+          io.readbytes(len)
+        end
+
+        def sensible_default
+          [ ]
+        end
+
+        def value_to_binary_string(val)
+          clamp_to_length(val).pack('C*')
+        end
+
+        class NdrFixedByteArrayArgProcessor < BinData::BaseArgProcessor
+          def sanitize_parameters!(obj_class, obj_params)
+            obj_params.must_be_integer(:length, :pad_byte)
+            obj_params.sanitize(:pad_byte) { |byte| sanitized_pad_byte(byte) }
+          end
+
+          private
+
+          def sanitized_pad_byte(byte)
+            if byte.is_a?(String)
+              raise ArgumentError, ':pad_byte must not contain more than 1 byte' if byte.bytesize > 1
+
+              byte = byte.ord
+            end
+            raise ArgumentError, ':pad_byte must be within the range of 0 - 255' unless ((byte >= 0) && (byte <= 255))
+
+            byte
+          end
+        end
+      end
+
       # An NDR Context Handle representation as defined in
       # [IDL Data Type Declarations - Basic Type Declarations](http://pubs.opengroup.org/onlinepubs/9629399/apdxn.htm#tagcjh_34_01)
       class NdrContextHandle < BinData::Primitive
--- a/lib/ruby_smb/dcerpc/ndr.rb
+++ b/lib/ruby_smb/dcerpc/ndr.rb
@@ -98,9 +98,16 @@ module RubySMB
       # An NDR Uni-dimensional Fixed Array of bytes representation as defined in:
       # [Transfer Syntax NDR - NDR Constructed Types](https://pubs.opengroup.org/onlinepubs/9629399/chap14.htm#tagcjh_19_03_03_01)
       class NdrFixedByteArray < BinData::BasePrimitive
-        optional_parameters :length, :pad_byte, :pad_front
+        optional_parameters :read_length, :length, :pad_byte, :pad_front
         default_parameters pad_byte: 0
         mutually_exclusive_parameters :length, :value
+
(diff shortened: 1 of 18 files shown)


wvu-r7:
This looks GREAT!

lib/ruby_smb/dcerpc/netlogon/netr_server_authenticate3_response.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Dcerpc
+    module Netlogon
+
+      # [3.5.4.4.2 NetrServerAuthenticate3 (Opnum 26)](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/3a9ed16f-8014-45ae-80af-c0ecb06e2db9)
+      class NetrServerAuthenticate3Response < BinData::Record
+        attr_reader :opnum
+
+        endian :little
+
+        netlogon_credential :server_credential
```
cdelafuente-r7, line 13: Same comment about the "possible" pointer to a NETLOGON_CREDENTIAL structure. This will need to be tested.

lib/ruby_smb/dcerpc/netlogon/netr_server_authenticate3_response.rb
```diff
@@ -0,0 +1,26 @@
...
+        uint32              :negotiate_flags
+        uint32              :account_rid
+        uint32              :status
```
cdelafuente-r7, line 16: Not a blocker, but all the other RPC structures in `ruby_smb` use `error_status` instead of `status`. I think it would be interesting to follow the same name convention.

lib/ruby_smb/dcerpc/netlogon/netr_server_req_challenge_request.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Dcerpc
+    module Netlogon
+
+      # [3.5.4.4.1 NetrServerReqChallenge (Opnum 4)](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/5ad9db9f-7441-4ce5-8c7b-7b771e243d32)
+      class NetrServerReqChallengeRequest < BinData::Record
+        attr_reader :opnum
+
+        endian :little
+
+        ndr_lp_str          :primary_name
```
cdelafuente-r7, line 13: Same comment about using names that match the documentation:
```suggestion
        logon_srv_handle    :primary_name
```

lib/ruby_smb/dcerpc/netlogon/netr_server_req_challenge_request.rb
```diff
@@ -0,0 +1,25 @@
...
+        ndr_string          :computer_name
+        netlogon_credential :client_challenge
```
cdelafuente-r7, line 15: Same comment about the "possible" pointer to a NETLOGON_CREDENTIAL structure. This will need to be tested too.

lib/ruby_smb/dcerpc/netlogon/netr_server_req_challenge_response.rb
```diff
@@ -0,0 +1,24 @@
...
+    module Netlogon
+
+      # [3.5.4.4.1 NetrServerReqChallenge (Opnum 4)](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/5ad9db9f-7441-4ce5-8c7b-7b771e243d32)
+      class NetrServerReqChallengeResponse < BinData::Record
+        attr_reader :opnum
+
+        endian :little
+
+        netlogon_credential :server_challenge
+        uint32              :status
```
cdelafuente-r7, line 14: Same here :)

cdelafuente-r7 (commented):
Thanks for adding this, @zeroSteiner! This is awesome! I left a few comments for you to review when you get a chance.

Also, for the implementation to be used by RubySMB client, some other changes are needed:
- The `Netlogon` requests need to be added to the `stub` field in https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/dcerpc/request.rb
- `netlogon` endpoint should be added to both https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb1/pipe.rb and https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb2/pipe.rb (`#initialize`).

zeroSteiner:
Alright at this point all feedback should be addressed.

* The Netlogon requests have been added to the stub field
* The Netlogon endpoint has been added to both the SMB1 and SMB2 `Pipe` classes
* Rspec unit tests have been added for each the request / response definitions (what's currently missing are the 2 crypto helper functions and the new data-types)

Regarding all of the implementation / documentation discrepancies, I don't really know what to say other than that each field is used as part of the exploit and appears to be working. Between the REMOVE and RESTORE actions, all of the request and most of the response fields are used.

lib/ruby_smb/dcerpc/ndr.rb
```diff
@@ -92,6 +95,85 @@ def set(v)
...
+          val
+        end
+
+        def fixed_byte_array(val)
+          val = val.bytes if val.is_a? String
+          val.to_ary
+        end
+
+        def read_and_return_value(io)
+          len = eval_parameter(:length) || 0
```
cdelafuente-r7, line 144: If `:length` parameter is not provided, nothing will be read. This can be confusing for the user since this parameter is not mandatory. Maybe `:length` could be set as a `mandatory_parameter`?

cdelafuente-r7 (commented):
Thanks for addressing these feedback! I just left one last comment about the `length` parameter in `NdrFixedByteArray`.

lib/ruby_smb/dcerpc/ndr.rb
zeroSteiner: That makes sense. In commit 013c9ed1ae654f96a1a1f7dc8bc24ffb06c0dc0b I implemented a `WarnNoReadLengthPlugin` so this behaves just like the `BinData::String` [class](https://github.com/dmendel/bindata/blob/5a523e2d826763a7505d73920de1a09ffdb4ef23/lib/bindata/string.rb) from which it is derived. When a value is specified without a `:read_length` a warning will be thrown. If neither is specified, then it's treated as an empty array just like the related string class is treated as an empty string.

cdelafuente-r7: Great! Sounds good to me! Thanks!

cdelafuente-r7 (approved):
Everything looks good to me. I will test this with https://github.com/rapid7/metasploit-framework/pull/14151, which leverages this code.

cdelafuente-r7:
Tested successfully against Windows Server 2016 x64 (see [my notes](https://github.com/rapid7/metasploit-framework/pull/14151#issuecomment-696281913) in the exploit module).

I will go ahead and land it shortly. Thanks for these great additions @zeroSteiner !

bowojori7:
Hello!

Sorry for asking such a noob question in what seems like an expert space. 
How do I install the modules on my Metasploit to take advantage of this exploit? 

Thank you.

cdelafuente-r7:
Hello @bowojori7, this module has been merged into Metasploit master branch and is available in the latest development version (see [this PR](https://github.com/rapid7/metasploit-framework/pull/14151) for details).

Please, avoid commenting on a closed PR. For general questions, it is preferred to use IRC (#metasploit on Freenode) or our Slack channel ([here](https://join.slack.com/t/metasploit/shared_invite/enQtOTMxMjg5ODY3NDYxLTYyZTVlNDM1ZTI3ZGQ0YTQ0NmI1YjgxMGEwYzFhOWNiZWQ4NDNmNTE1MWNiYzJhMzgwMGY0ZWY2NTY0MjA2M2I) to join).

PR #167, opened 2021-02-08
Add github actions for testing

adfoster-r7:
Adding github actions for testing:

image

Note that we can't add ruby 3.0.0 to the test matrix as it's currently broken by a bug in ruby hashes https://bugs.ruby-lang.org/issues/17488

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1,0 +1,56 @@
+name: Verify
+
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  test:
+    runs-on: ubuntu-16.04
+    timeout-minutes: 40
+
+    strategy:
+      fail-fast: true
+      matrix:
+        ruby:
+          - 2.5
+          - 2.6
+          - 2.7
+        test_cmd:
+          - bundle exec rspec
+
+    env:
+      RAILS_ENV: test
+
+    name: Ruby ${{ matrix.ruby }} - ${{ matrix.test_cmd }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+
+      - name: Setup bundler
+        run: |
+          gem install bundler
+      - uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+      - name: Bundle install
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+      - name: ${{ matrix.test_cmd }}
+        run: |
+          echo "${CMD}"
+          bash -c "${CMD}"
+        env:
+          CMD: ${{ matrix.test_cmd }}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,0 @@
-language: ruby
-sudo: false
-rvm:
-  - '2.5.8'
-  - '2.6.6'
-  - '2.7.0'
--- a/README.md
+++ b/README.md
@@ -1,6 +1,5 @@
 # RubySMB
 
-[![Build Status](https://travis-ci.org/rapid7/ruby_smb.svg?branch=master)](https://travis-ci.org/rapid7/ruby_smb)
 [![Code Climate](https://codeclimate.com/github/rapid7/ruby_smb.png)](https://codeclimate.com/github/rapid7/ruby_smb)
 [![Coverage Status](https://coveralls.io/repos/github/rapid7/ruby_smb/badge.svg?branch=master)](https://coveralls.io/github/rapid7/ruby_smb?branch=master)
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -7,7 +7,7 @@ end
 require 'coveralls'
 require 'ruby_smb'
 
-if ENV['TRAVIS'] == 'true'
+if ENV['CI'] == 'true'
   # don't generate local report as it is inaccessible on travis-ci, which is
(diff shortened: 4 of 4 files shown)


cdelafuente-r7 (approved):
I'm not too familiar with Github actions, but it looks good to me! Thanks for updating this.

Also, I'm curious about this bug in ruby 3.0.0. Is RubySMB using `DelegateClass`? Or maybe `BinData`?

adfoster-r7:
I don't think the problem was specifically with DelegateClass, but the hash implementation itself

Here's the thread that I had posted internally detailing my debug process:

---

Looks like the ruby_smb gem fails occasionally on Ruby 3.0:
```
  1) RubySMB::Dcerpc::Request#stub with a Winreg endpoint selects the expected packet structure
     Failure/Error: expect(packet.stub).to eq(RubySMB::Dcerpc::Winreg::OpenRootKeyRequest.new(opnum: opnum))

       expected: {:p_registry_server_name=>:null, :sam_desired=>{:reserved=>0, :key_create_link=>0, :key_notify=>0, :k...ic_write=>0, :generic_execute=>0, :generic_all=>0, :reserved4=>0, :maximum=>1, :system_security=>0}}
            got: ""
```

Normally non-deterministic rspec tests are a sign of a flakey test, but it’s been pretty solid on Ruby 2.7. I even ran with rspec’s [--bisect](https://relishapp.com/rspec/rspec-core/docs/command-line/bisect) flag to confirm it wasn’t an ordering/state issue

The test it fails on:
https://github.com/rapid7/ruby_smb/blob/09b3160d5dd4a0ecee3fac89b66eab50170bb334/spec/lib/ruby_smb/dcerpc/request_spec.rb#L103-L105

It looks like the bindata choice occasionally does the wrong thing:

https://github.com/rapid7/ruby_smb/blob/09b3160d5dd4a0ecee3fac89b66eab50170bb334/lib/ruby_smb/dcerpc/request.rb#L15-L33

After putting in a breakpoint it seems that opnum is a BinData object, not an integer:
```
[20] pry(#<BinData::SanitizedChoices>)> opnum
=> 0
[21] pry(#<BinData::SanitizedChoices>)> opnum.class
=> BinData::Uint16le
```

Digging into how BinData implements the choice functionality, it seems to have an internal hash for keeping track of choices:
```
From: /Users/user/.rvm/gems/ruby-3.0.0/gems/bindata-2.4.8/lib/bindata/sanitize.rb:152 BinData::SanitizedChoices#[]:

    151: def [](key)
    152:   @choices[key]
    153: end
The hash uses integer keys for its lookup:
[18] pry(#<BinData::SanitizedChoices>)> @choices.keys
=> [0, 1, 2, 3, 4, 27, 32, 33, 5, 9, 10, 15, 16, 17, 6, 20]
[19] pry(#<BinData::SanitizedChoices>)> @choices.keys.map(&:class)
=> [Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer, Integer]
The key lookup is opnum , which as I spotted was actually a BinData object:
[20] pry(#<BinData::SanitizedChoices>)> key
=> 0
[21] pry(#<BinData::SanitizedChoices>)> key.class
=> BinData::Uint16le
```

A fix could be:
```diff
- choice 'Winreg', selection: -> { opnum } do
+ choice 'Winreg', selection: -> { opnum.to_i } do
```

I’m tempted to say that BinData should be changed to be using key.to_i somewhere, but it’s weird that it breaks only on Ruby 3.0

I dug deeper, it seems like the BinData object quacks like an integer, but fails lookup for some reason:
```
[25] pry(#<BinData::SanitizedChoices>)> key == 0.to_i
=> true
[25] pry(#<BinData::SanitizedChoices>)> key === 0.to_i
=> true
[26] pry(#<BinData::SanitizedChoices>)> @choices.keys[0] == key
=> true
[28] pry(#<BinData::SanitizedChoices>)> @choices[key]
=> nil
[43] pry(#<BinData::SanitizedChoices>)> @choices[key.to_i].class
=> BinData::SanitizedPrototype
I was thinking weird hashing differences, but nope:
[8] pry(#<BinData::SanitizedChoices>)> key.hash
=> 3007275666221366699
[9] pry(#<BinData::SanitizedChoices>)> 0.hash
=> 3007275666221366699
[10] pry(#<BinData::SanitizedChoices>)> @choices.keys.map(&:hash)
=> [3007275666221366699, 977401519670554717, 2416423894636754036, 387683318277990506, 1826704593738853376, 536335832567336451, 3947708622480110334, 766046541512406665, 4417522539582447671, 3237232924370342128, 5562082199671962, 466934854939131662, 2895825195904021269, 876364532024555299, 1235852796394923053, 2877464285071549434]
```

The bindata object overrides the expected methods:

```
[5] pry(#<BinData::SanitizedChoices>)> key.method(:==).source_location
=> ["/Users/adfoster/.rvm/gems/ruby-3.0.0/gems/bindata-2.4.8/lib/bindata/base.rb", 231]
[6] pry(#<BinData::SanitizedChoices>)> key.method(:eql?).source_location
=> ["/Users/adfoster/.rvm/gems/ruby-3.0.0/gems/bindata-2.4.8/lib/bindata/base_primitive.rb", 119]
[7] pry(#<BinData::SanitizedChoices>)> 0.eql?(key)
=> false
[8] pry(#<BinData::SanitizedChoices>)> key.eql?(0)
=> true
```

After hitting dead ends, I searched for Ruby 3.0 hash regression issues and this looked super similar, it’s created a month ago https://bugs.ruby-lang.org/issues/17488

Looks like there was a fix landed:
https://github.com/ruby/ruby/commit/b2beb8586e930c168af434d6545f75d76123192b

I've now built the the custom ruby commit locally with rvm:
```
rvm install 3.0-testing-hash --sha b2beb8586e930c168af434d6545f75d76123192b
```

I was able to confirm locally that ruby_smb works again with the custom build, and passes 100 runs in a row, versus previously failing >50% of the time:

```
for run in {1..100}; do bundle exec rspec './spec/lib/ruby_smb/dcerpc/request_spec.rb[1:14:3:2]'; done
```

PR #169, opened 2021-04-07
Improve Compression Capabilities

zeroSteiner:
This makes a few changes to allow RubySMB to be used for exploiting CVE-2020-0796 (SMBGhost). The changes include:

* Adding a `#compressed_data` field to the `CompressionTransformHeader` so the payload can be included which is required for sending this data. This mirrors how the `TransformHeader` works with the `#encrypted_data`, which is the same idea (the transformed payload needs to be included.
* Added an LZNT1 algorithm implementation along with some basic unit tests that get decent coverage. The implementation was ported to Ruby from [this Python implementation](https://github.com/you0708/lznt1).
    * The specs are intended to test blocks that are copied as-is and blocks that are compressed. They should help us identify errors we might introduce into the algorithm later on since the inputs and outputs should be deterministic.
* Fixed a whole bunch of typos I found while reading through RubySMB and working on SMBGhost.

## Testing
This will be required by the exploit that I'll be submitting to Metasploit and can be partially tested in conjunction with that. The LZNT1 compression capabilities are required as is the `#compressed_data` field.

- [ ] Make sure the unit tests pass and the code looks good, from RubySMB's perspective it's all dead code

### DIFF
--- a/lib/ruby_smb/compression/lznt1.rb
+++ b/lib/ruby_smb/compression/lznt1.rb
@@ -8,10 +8,10 @@ module RubySMB
           compressed = compress_chunk(chunk)
           # chunk is compressed
           if compressed.length < chunk.length
-            out << [ 0xb000 | (compressed.length - 1) ].pack('v')
+            out << [0xb000 | (compressed.length - 1)].pack('v')
             out << compressed
           else
-            out << [ 0x3000 | (chunk.length - 1) ].pack('v')
+            out << [0x3000 | (chunk.length - 1)].pack('v')
             out << chunk
           end
           buf = buf[chunk_size..-1]
@@ -72,16 +72,16 @@ module RubySMB
       def self.decompress(buf, length_check: true)
         out = ''
         until buf.empty?
-          header = buf[0..2].unpack('v').first
+          header = buf[0..2].unpack1('v')
           length = (header & 0xfff) + 1
           raise EncodingError, 'invalid chunk length' if length_check && length > (buf.length - 2)
 
           chunk = buf[2...length + 2]
-          if header & 0x8000 == 0
-            out << chunk
-          else
-            out << decompress_chunk(chunk)
-          end
+          out << if header & 0x8000 == 0
+                   chunk
+                 else
+                   decompress_chunk(chunk)
+                 end
           buf = buf[length + 2..-1]
         end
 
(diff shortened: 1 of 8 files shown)


lib/ruby_smb/smb2/packet/compression_transform_header.rb
```diff
@@ -11,6 +11,7 @@ class CompressionTransformHeader < BinData::Record
         uint16           :compression_algorithm,            label: 'Compression Algorithm'
         uint16           :flags,                            label: 'Flags'
         uint32           :offset,                           label: 'Offset / Length'
+        array            :compressed_data,                  label: 'Compressed Data', type: :uint8, read_until: :eof
       end
```
cdelafuente-r7, line 15: Not a blocker, just a note here. It looks like the `COMPRESSION_TRANSFORM_HEADER` protocol definition has been updated and is now `SMB2_COMPRESSION_TRANSFORM_HEADER_CHAINED`and `SMB2_COMPRESSION_TRANSFORM_HEADER_UNCHAINED` (see https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/1d435f21-9a21-4f4c-828e-624a176cf2a0).

It is likely out of scope for this PR, but it is something we will need to update when implementing compression capability in RubySMB client.

smcintyre-r7: Wow yeah literally updated just a couple of days ago. Good catch! We can add a note that this structure is here for historical reasons because there was a point in time where it was defined as is.

cdelafuente-r7: Absolutely, a note would be certainly useful. Thanks!

cdelafuente-r7 (commented):
Thanks for adding the first compression algorithm to RubySMB @smcintyre-r7 ! I just left a few comments regarding issues I found while I was testing against a real Windows host. Specifically, the issue with the logic that checks if the chunk is empty in `decompress_chunk` was breaking the execution.

lib/ruby_smb/compression/lznt1.rb
```diff
@@ -0,0 +1,165 @@
...
+          out << tmp
+        end
+
+        out
+      end
+
+      def self.decompress(buf, length_check: true)
+        out = ''
+        until buf.empty?
+          header = buf[0..2].unpack1('v')
```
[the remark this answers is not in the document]
zeroSteiner: Yeah, I recently learned that Ruby is permissive and will just ignore all unused bytes, so I'll reduce this down to just `buf.unpack1('v')`. Python on the other hand requires the number of bytes to match exactly 🤷🏻 .

lib/ruby_smb/compression/lznt1.rb
```diff
@@ -0,0 +1,165 @@
...
+                out_offset = out[-offset..-1]
+                tmp = out_offset * (0xfff / out_offset.length + 1)
+                out << tmp[0...length]
+              else
+                out << out[-offset..-offset + length - 1]
+              end
+              chunk = chunk[2..-1]
+
+              break if chunk.empty?
+            end
```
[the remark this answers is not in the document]
zeroSteiner: Great catch!

PR #171, opened 2021-05-10
Adds ntlm_flags & NegotiateResponseExtended

agalway-r7:
Required for [eternalblue_win8 port to ruby](https://github.com/rapid7/metasploit-framework/pull/15182).

Extends `Net::NTLM::Client` initializer to include `ntlm_flags`. 

Also allows `negotiate_response` to parse returning `NegotiateResponseExtended` packets into `RubySMB::SMB1::Packet::NegotiateResponseExtended`. This is a useful feature for the entire library, not just this one PR.

### DIFF
--- a/lib/ruby_smb/client.rb
+++ b/lib/ruby_smb/client.rb
@@ -277,7 +277,8 @@ module RubySMB
     # @param smb1 [Boolean] whether or not to enable SMB1 support
     # @param smb2 [Boolean] whether or not to enable SMB2 support
     # @param smb3 [Boolean] whether or not to enable SMB3 support
-    def initialize(dispatcher, smb1: true, smb2: true, smb3: true, username:, password:, domain: '.', local_workstation: 'WORKSTATION', always_encrypt: true)
+    def initialize(dispatcher, smb1: true, smb2: true, smb3: true, username:, password:, domain: '.',
+                   local_workstation: 'WORKSTATION', always_encrypt: true, ntlm_flags: default_flags)
       raise ArgumentError, 'No Dispatcher provided' unless dispatcher.is_a? RubySMB::Dispatcher::Base
       if smb1 == false && smb2 == false && smb3 == false
         raise ArgumentError, 'You must enable at least one Protocol'
@@ -306,18 +307,12 @@ module RubySMB
       # SMB 3.x options
       @session_encrypt_data = always_encrypt
 
-      negotiate_version_flag = 0x02000000
-      flags = Net::NTLM::Client::DEFAULT_FLAGS |
-        Net::NTLM::FLAGS[:TARGET_INFO] |
-        negotiate_version_flag ^
-        Net::NTLM::FLAGS[:OEM]
-
       @ntlm_client = Net::NTLM::Client.new(
         @username,
         @password,
         workstation: @local_workstation,
         domain: @domain,
-        flags: flags
+        ntlm_flags: ntlm_flags
       )
 
       @tree_connects = []
@@ -368,31 +363,28 @@ module RubySMB
 
     # Performs protocol negotiation and session setup. It defaults to using
     # the credentials supplied during initialization, but can take a new set of credentials if needed.
-    def login(username: self.username, password: self.password, domain: self.domain, local_workstation: self.local_workstation)
+    def login(username: self.username, password: self.password,
+              domain: self.domain, local_workstation: self.local_workstation,
+              ntlm_flags: default_flags)
       negotiate
-      session_setup(username, password, domain, true,
-                    local_workstation: local_workstation)
+      session_setup(username, password, domain,
+                    local_workstation: local_workstation,
+                    ntlm_flags: ntlm_flags)
     end
 
     def session_setup(user, pass, domain, do_recv=true,
-                      local_workstation: self.local_workstation)
+                      local_workstation: self.local_workstation, ntlm_flags: default_flags)
       @domain            = domain
       @local_workstation = local_workstation
       @password          = pass.encode('utf-8') || ''.encode('utf-8')
       @username          = user.encode('utf-8') || ''.encode('utf-8')
-
-      negotiate_version_flag = 0x02000000
-      flags = Net::NTLM::Client::DEFAULT_FLAGS |
-        Net::NTLM::FLAGS[:TARGET_INFO] |
-        negotiate_version_flag ^
-        Net::NTLM::FLAGS[:OEM]
 
       @ntlm_client = Net::NTLM::Client.new(
           @username,
           @password,
           workstation: @local_workstation,
           domain: @domain,
-          flags: flags
+          ntlm_flags: ntlm_flags
       )
 
       authenticate
@@ -653,6 +645,18 @@ module RubySMB
         @preauth_integrity_hash_algorithm,
         @preauth_integrity_hash_value + data.to_binary_s
       )
+    end
+
+    private
+
+    def default_flags
+      negotiate_version_flag = 0x02000000
+      flags = Net::NTLM::Client::DEFAULT_FLAGS |
+        Net::NTLM::FLAGS[:TARGET_INFO] |
+        negotiate_version_flag ^
+        Net::NTLM::FLAGS[:OEM]
+
+      flags
     end
   end
 end
--- a/lib/ruby_smb/client.rb
+++ b/lib/ruby_smb/client.rb
@@ -312,7 +312,7 @@ module RubySMB
         @password,
         workstation: @local_workstation,
         domain: @domain,
-        ntlm_flags: ntlm_flags
+        flags: ntlm_flags
       )
 
       @tree_connects = []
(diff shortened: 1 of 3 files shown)


lib/ruby_smb/client/negotiation.rb
```diff
@@ -57,15 +57,19 @@ def negotiate_request
...
         response = nil
         if smb1
-          packet = RubySMB::SMB1::Packet::NegotiateResponseExtended.read raw_data
+          packet = RubySMB::SMB1::Packet::NegotiateResponse.read raw_data
+
+          if packet.smb_header.flags2.extended_security == 1
+            packet = RubySMB::SMB1::Packet::NegotiateResponseExtended.read raw_data
+          end
+
           response = packet if packet.valid?
```
cdelafuente-r7, line 73: Apparently,  the [extended security capability](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb/d367854f-5eee-45e8-a588-eed596a1a521) flag needs to be checked in the server response instead of the SMB header flag2 `extended_security` flag. This is what the [`valid?`](https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/smb1/packet/negotiate_response_extended.rb#L41) method do.

I also believe it is more likely to receive an `NegotiateResponseExtended` response than a `NegotiateResponse` response. So, an improvement would be to parse with the `NegotiateResponseExtended` structure first and call `valid?`, which is what the original code do. Then, if it is not valid, parse with the `NegotiateResponse` structure. This logic can be added to the `if response.nil?` block below.

lib/ruby_smb/client.rb
```diff
@@ -306,18 +307,12 @@ def initialize(dispatcher, smb1: true, smb2: true, smb3: true, username:, passwo
...
-        negotiate_version_flag ^
-        Net::NTLM::FLAGS[:OEM]
-
       @ntlm_client = Net::NTLM::Client.new(
         @username,
         @password,
         workstation: @local_workstation,
         domain: @domain,
-        flags: flags
+        ntlm_flags: ntlm_flags
```
cdelafuente-r7, line 315: I might be missing something here, but I cannot find the `ntlm_flags` key word argument being used in the `Net::NTLM::Client` class [constructor](https://github.com/WinRb/rubyntlm/blob/master/lib/net/ntlm/client.rb#L21-L27).

lib/ruby_smb/client.rb
```diff
@@ -368,31 +363,28 @@ def increment_smb_message_id(packet)
...
-        negotiate_version_flag ^
-        Net::NTLM::FLAGS[:OEM]
-
       @ntlm_client = Net::NTLM::Client.new(
           @username,
           @password,
           workstation: @local_workstation,
           domain: @domain,
-          flags: flags
+          ntlm_flags: ntlm_flags
```
cdelafuente-r7, line 387: Same here, it looks like `ntlm_flags` key word argument is not used in the `Net::NTLM::Client` class constructor.

agalway-r7: Yeah this and the above `ntlm_flag` both broke existing implementation and needed to be removed 👍

lib/ruby_smb/client/negotiation.rb
```diff
@@ -66,6 +65,11 @@ def negotiate_response(raw_data)
         response = nil
         if smb1
           packet = RubySMB::SMB1::Packet::NegotiateResponseExtended.read raw_data
+
+          unless packet.valid?
+            packet = RubySMB::SMB1::Packet::NegotiateResponse.read raw_data
+          end
+
```
dwelch-r7, line 72: what's this doing? if the packet isn't valid you read it again? seems a bit weird is this common?

agalway-r7: No worries, I made that mistake a bunch when I started on this 😄

spec/lib/ruby_smb/client_spec.rb
```diff
@@ -1015,12 +1015,6 @@
           bogus_response.smb_header.command = 0xff
           expect { smb1_client.negotiate_response(bogus_response.to_binary_s) }.to raise_error(RubySMB::Error::InvalidPacket)
         end
-
-        it 'considers the response invalid if Extended Security is not enabled' do
-          bogus_response = smb1_extended_response
-          bogus_response.parameter_block.capabilities.extended_security = 0
-          expect { smb1_client.negotiate_response(bogus_response.to_binary_s) }.to raise_error(RubySMB::Error::InvalidPacket)
-        end
```
adfoster-r7, line 1023: Instead of deleting the test, is it worth updating it to reflect the new implementation? 👀

lib/ruby_smb/client/negotiation.rb
agalway-r7: Cheers man 👍

lib/ruby_smb/client/negotiation.rb
```diff
@@ -74,10 +78,10 @@ def negotiate_response(raw_data)
...
         if response.nil?
           if packet.packet_smb_version == 'SMB1'
-            extended_security = if packet.is_a? RubySMB::SMB1::Packet::NegotiateResponseExtended
-              packet.parameter_block.capabilities.extended_security
+            extended_security = if packet.parameter_block['capabilities'] && packet.parameter_block['capabilities']['extended_security']
+              packet.parameter_block['capabilities']['extended_security']
             else
-              "n/a"
+              0
             end
```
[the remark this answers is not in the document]
agalway-r7: You're right, there's no need to pull out the `extended_security` fields when it's only a `NegotiateResponse`  packet that can reach this code block.

spec/lib/ruby_smb/client_spec.rb
agalway-r7: https://youtu.be/5r06heQ5HsI

lib/ruby_smb/client/negotiation.rb
```diff
@@ -57,7 +57,6 @@ def negotiate_request
 
       # Takes the raw response data from the server and tries
       # parse it into a valid Response packet object.
-      # This method currently assumes that all SMB1 will use Extended Security.
       #
       # @param raw_data [String] the raw binary response from the server
       # @return [RubySMB::SMB1::Packet::NegotiateResponseExtended] when the response is an SMB1 Extended Security Negotiate Response Packet
```
adfoster-r7, line 62: I think we're missing another return comment here now

lib/ruby_smb/client/negotiation.rb
```diff
@@ -66,6 +65,11 @@ def negotiate_response(raw_data)
         response = nil
         if smb1
           packet = RubySMB::SMB1::Packet::NegotiateResponseExtended.read raw_data
+
+          unless packet.valid?
+            packet = RubySMB::SMB1::Packet::NegotiateResponse.read raw_data
+          end
```
adfoster-r7, line 71: Just confirming if this is secure to do. I don't know enough about SMB, but if we've negotiated some sort of extended security session, it feels like falling back here feels like it should be predicated on the whether or not extended security was enabled in the first place? 👀

agalway-r7: Agreed, but that may be outside the scope of this PR.

lib/ruby_smb/client/negotiation.rb
agalway-r7: 👍

lib/ruby_smb/client/negotiation.rb
adfoster-r7: Again, I don't know the scope of this change - but if the PR adds a potential security regression, then it's definitely in the scope of this PR 😄 

cc @cdelafuente-r7 I assume there's no negative repercussions to this line? 👀

cdelafuente-r7: The client needs to follow what the server claims to support. At this point, the client has sent a `NegotiateRequest` asking for [Extended Security](https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/client/negotiation.rb#L213) and the server should answer with the `CAP_EXTENDED_SECURITY` capability flag, meaning it supports Extended Security and the response contains the expected Security Blob (GSS protocol negotiation). Parsing first with a `NegotiateResponseExtended` structure will just do that. If the `CAP_EXTENDED_SECURITY` capability flag is not set, `packet.valid?` will fail and the client will fail back to a normal `NegotiateResponse`. This process is described [here](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb/d367854f-5eee-45e8-a588-eed596a1a521).

That said, The Security Blob is not used this way, for simplicity. RubySMB assumes the server uses NTLMSSP as authentication mechanism and generates the corresponding Security Blob in the upcoming Session Setup request, which follows the negotiation step. It ignores the `NegotiateResponseExtended` Security Blob (GSS protocol negotiation). RubySMB does not support other authentication mechanism anyway. So, if the server does not support NTLMSSP or any Extended Security related protocol at all, the authentication is likely to fail.

PR #174, opened 2021-06-29
Adds support for smb1 negotiate response packets

agalway-r7:
Extends the ruby_smb client to include a `challenge` field, which is populated by an additional case statement in the `parse_negotiate_response` method. An extension to the `negotiate` API to allow callers to pass in their own negotiate SMB packet is also added.


adfoster-r7:
PR closed as it's not from a fork :+1:

PR #176, opened 2021-06-30
Adds support for smb1 negotiate response packets

agalway-r7:
Extends the ruby_smb client to include a `challenge` field, which is populated by an additional case statement in the `parse_negotiate_response` method. An extension to the `negotiate` API to allow callers to pass in their own negotiate SMB packet is also added.

### DIFF
--- a/lib/ruby_smb/client.rb
+++ b/lib/ruby_smb/client.rb
@@ -138,6 +138,11 @@ module RubySMB
     #   @return [Boolean]
     attr_accessor :smb1
 
+    # Challenge value returned by the server during SMB1 negotiation with Extended Security turned off
+    # @!attribute [rw] challenge
+    #       @return [String]
+    attr_accessor :challenge
+
     # Whether or not the Client should support SMB2
     # @!attribute [rw] smb2
     #   @return [Boolean]
--- a/lib/ruby_smb/client/negotiation.rb
+++ b/lib/ruby_smb/client/negotiation.rb
@@ -9,8 +9,8 @@ module RubySMB
       # It also keeps track of the negotiated dialect.
       #
       # @return [void]
-      def negotiate
-        request_packet  = negotiate_request
+      def negotiate(negotiate_packet: nil)
+        request_packet  = negotiate_packet ? negotiate_packet : negotiate_request
         raw_response    = send_recv(request_packet)
         response_packet = negotiate_response(raw_response)
         # The list of dialect identifiers sent to the server is stored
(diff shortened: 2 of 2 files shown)


lib/ruby_smb/client.rb
```diff
@@ -138,6 +138,11 @@ class Client
     #   @return [Boolean]
     attr_accessor :smb1
 
+    # Challenge value returned by the server during SMB1 negotiation with Extended Security turned off
+    # @!attribute [rw] challenge
+    #       @return [String]
+    attr_accessor :challenge
```
cdelafuente-r7, line 144: Is `challenge` something that is only possible for SMB1 with Extended Security turned off? I have the feeling that `challenge` also exists for SMB1 with Ext. Sec, SMB2 and SMB3. If it is, it might be a good idea to add support for every case scenario and give access to the `challenge` in any situation.

lib/ruby_smb/client/negotiation.rb
```diff
@@ -9,8 +9,8 @@ module Negotiation
       # It also keeps track of the negotiated dialect.
       #
       # @return [void]
-      def negotiate
-        request_packet  = negotiate_request
+      def negotiate(negotiate_packet: nil)
```
cdelafuente-r7, line 12: I’m not sure about adding `negotiate_packet` argument to `#negotiate`. It might be confusing since it is only used for a very limited use case. Did you consider moving this logic directly to the module you’re working on?
For example, I think it is the only part you’ll need:
```ruby
raw_response    = client.send_recv(custom_negotiate_packet)
response_packet = client.negotiate_response(raw_response)
response_packet.dialects = request_packet.dialects
# parse_negotiate_response used section:
client.smb1_response(response_packet)
challenge = response_packet.data_block.challenge
```

cdelafuente-r7 (commented):
Thanks for this PR @agalway-r7 ! I left a few comments for you to review. Please, let me know if you have any questions.

lib/ruby_smb/client/negotiation.rb
agalway-r7: Yeah this is a nicer way of doing it will change over 👍

agalway-r7:
Closed in favour of making the changes locally in the [eternalblue module](https://github.com/rapid7/metasploit-framework/pull/15217).

This is done because the need for a custom negotiation packet & supporting `NegotiateResponse` is more or less exclusive to NTLM SMB1 requests, whicha re pretty unlikely to come up in the future.

PR #177, opened 2021-08-06
Add an initial SMB server implementation

zeroSteiner:
ImTheSMBServerNow

# SMB Server
This PR adds an initial implementation of an SMB server powered by Ruby SMB. This partial implementation supports SMB version 1 (dialect `NT LM 0.12`) through SMB version 3.1.1 (the latest as of this writing) and every dialect in the 2.x and 3.x families. Everything needed to fully authenticate a user is included (but nothing more). This includes the ability to perform protocol negotiation and session setup.

At this point the server should provide everything necessary to run authentication based attacks like capturing and relaying. Things like hosting files will be done in another round of work.

## Dialect Module
Folks familiar with SMB know that the SMB versions 2 and 3 are very closely related. If you browse through the code base you'll notice that SMB3 related packets are defined under the `RubySMB::SMB2` module and *alot* of code is reused from the SMB2 logic to provide SMB3 functionality. I added a new dialect module to will translate a dialect from any of the versions into an object containing useful metadata that can be used by us developers to group the dialect using varying degrees of specificity from the exact dialect, to the version, to the family. This also allows code that needs to differentiate between version 1 and version 2 or 3 to simply check the `#order` attribute. In this context "order" is used because it's above the family in the [Taxonomic Rank](https://en.wikipedia.org/wiki/Taxonomic_rank). This allows us to be consistent about calling "SMB 2.x" and "SMB 3.x" a family which is aligned with the Microsoft SMB specs.

These metadata objects also have a few name attributes that will be helpful to print in diagnostic output in the future to describe the dialect to users. See the `version_name` and `full_name` attributes, which again have varying degrees of specificity for use in different contexts.

## Authentication
Authentication is abstracted into a provider module which is server specific and generates an `Authenticator` object instance which performs the actual authentication process and is specific to a particular connection. This is important because the provider object is shared across connections and could become confused if it were to handle multiple simultaneous authentication processes.

There are base classes for both the provider and authenticator that do nothing but define the interface and can be used in the future.

Authentication providers can also specify that they permit anonymous authentication. As the name implies, this configuration allows clients to connect and "authenticate" to the server without specifying any credentials. Windows 10 does not permit this by default (probably because crypto keys can't be negotiated in this case which eliminates some security protections).

### NTLM Authentication
There is a full NTLM-backed authentication provider included in this implementation. This provider uses the standard NTLMSSP protocol and goes through the full 4-way handshake involving the transfer of Net-NTLMv1 and Net-NTLMv2 challenge and responses. The provider allows accounts to be defined and stored in memory, though a database or file backend could easily be added by a new class and overriding the get and put account methods.

Account usernames and domains are case-preserving but insensitive for comparisons. This means a user logging in as `workgroup\rubysmb` can authenticate just fine even if the account was defined as being `WORKGROUP\RubySMB` so long as the password is correct of course.

The NTLM messages are provided by the `Net::NTLM` module which RubySMB is already using for client-side authentication. One thing that could be problematic is that the NTLM messages are able to be in one of two encodings based on a flag within the message. Every client implementation I tested (see the Testing section below) uses UTF-16LE. A client implementation that **does not** use UTF-16LE may be unable to authenticate and is something that should be tested if such a client can be identified.

The provided NTLM provider class makes it very easy to manipulate the server challenge for nefarious purposes but defaults to using `SecureRandom` to properly generate a challenge. To override this behavior, a generator block can be provided:

```ruby
provider.generate_server_challenge do
  "\x11\x22\x33\x44\x55\x66\x77\x88"
end
```

or the method can be redefined as seen in the `example/auth_capture.rb` server.
```ruby
class HaxorNTLMProvider < RubySMB::Gss::Provider::NTLM
  def generate_server_challenge(&block)
    "\x11\x22\x33\x44\x55\x66\x77\x88"
  end
end
```

Since the `Net::NTLM` module is capable of treating NTLM hashes as passwords, an account can actually be provided using a valid NTLM hash as its password and this will work correctly.

## Prior Testing
image
This is an example of a Windows 10 client connection to a test harness, negotiating SMB version 3.1.1. The original authentication as the smcintyre user fails, and then Windows prompts for the username and password. At that point, I enter the credentials that are valid for the harness of `WORKGROUP\RubySMB` and `password` which allows it to authenticate and then attempt the TreeConnect request proving that the SMB 3 crypto keys were calculated correctly.

This implementation has been tested over the course of development using a number of clients including:

* RubySMB's `example/authenticate.rb`
* Windows
    * Windows XP
    * Windows 7
    * Windows 10
* Fedora's Nautilus file manager

In each case the client should be able to connect to, negotiate with and authenticate to the SMB server. Once authenticated, using Wireshark, it should be visible that a TreeConnect request is sent by the client showing that the authentication process was successful. Of particular note, SMB 3.1.1 needs to send a `TreeConnect` request because the Windows client will reject the server's response that authentication was successfully established if the response is not signed.

> If Connection.Dialect is "3.1.1", and if SMB2_FLAGS_SIGNED is not set in the Flags field of the SMB2 packet header of the response, the client MUST return an error to the calling application.

Source: [3.2.5.3.1 Handling a New Authentication](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/7fd079ca-17e6-4f02-8449-46b606ea289c)

Receiving the TreeConnect request shows that everything the server currently supports is working correctly, particularly the correct crypto keys for SMB 2 and SMB 3. After the TreeConnect request is received, it's expected that the server will crash and not handle that request at this time.

### Test Servers
This implementation includes a demonstration server that will capture authentication similar to Metasploit's `auxiliary/server/capture/smb` module. This example server is located at `example/auth_capture.rb` and simply logs the Net-NTLMv1 and Net-NTLMv2 challenge and response messages necessary for an attacker to perform an offline attack against the credentials of the user that initiated the connection.

<details>
<summary>Example Capture Output</summary>

The following is an example of the captured output where the client was the `example/authenticate.rb` script. The different versions of SMB that are being negotiated can be seen in the left hand column.

```
ruby auth_capture.rb
server is running
received connection
[SMB v3] NTLMv2-SSP Client   : 127.0.0.1
[SMB v3] NTLMv2-SSP Username : .\RubySMB
[SMB v3] NTLMv2-SSP Hash     : RubySMB::.:1122334455667788:9c538a23be908b316aac01868216c2a9:010100000000000080a54102fb8ad701351f9b4af704b8a600000000020012004c004f00430041004c0048004f0053005400010012004c004f00430041004c0048004f0053005400040002000000030012004c004f00430041004c0048004f00530054000700080080a54102fb8ad7010000000000000000
received connection
[SMB v2] NTLMv2-SSP Client   : 127.0.0.1
[SMB v2] NTLMv2-SSP Username : .\RubySMB
[SMB v2] NTLMv2-SSP Hash     : RubySMB::.:1122334455667788:5c12aec8b72e86c49a3bd5b395392fd5:010100000000000080a54102fb8ad70105feff0fcf247f7100000000020012004c004f00430041004c0048004f0053005400010012004c004f00430041004c0048004f0053005400040002000000030012004c004f00430041004c0048004f00530054000700080080a54102fb8ad7010000000000000000
received connection
[SMB v1] NTLMv2-SSP Client   : 127.0.0.1
[SMB v1] NTLMv2-SSP Username : .\RubySMB
[SMB v1] NTLMv2-SSP Hash     : RubySMB::.:1122334455667788:52d3fb5ea7d4d18fde57bc369dfebc94:010100000000000080a54102fb8ad701248b266e8b38ad1700000000020012004c004f00430041004c0048004f0053005400010012004c004f00430041004c0048004f0053005400040002000000030012004c004f00430041004c0048004f00530054000700080080a54102fb8ad7010000000000000000
received connection
[SMB v2] NTLMv2-SSP Client   : 127.0.0.1
[SMB v2] NTLMv2-SSP Username : .\RubySMB
[SMB v2] NTLMv2-SSP Hash     : RubySMB::.:1122334455667788:d1902819dd714910efd22ca0322edd9f:010100000000000080a54102fb8ad701362172730ac6bc0700000000020012004c004f00430041004c0048004f0053005400010012004c004f00430041004c0048004f0053005400040002000000030012004c004f00430041004c0048004f00530054000700080080a54102fb8ad7010000000000000000
received connection
[SMB v3] NTLMv2-SSP Client   : 127.0.0.1
[SMB v3] NTLMv2-SSP Username : .\RubySMB
[SMB v3] NTLMv2-SSP Hash     : RubySMB::.:1122334455667788:b92d74028ea93b670cae15177270ef30:010100000000000080a54102fb8ad7019240da74578df96e00000000020012004c004f00430041004c0048004f0053005400010012004c004f00430041004c0048004f0053005400040002000000030012004c004f00430041004c0048004f00530054000700080080a54102fb8ad7010000000000000000
```
</details>

This server is included but doesn't demonstrate a fully functional authenticator. To test the full NTLM authentication provider use a simple harness:

```ruby
#!/usr/bin/ruby

require 'bundler/setup'
require 'ruby_smb'
require 'ruby_smb/gss/provider/ntlm'

ntlm_provider = RubySMB::Gss::Provider::NTLM.new(allow_anonymous: true)
ntlm_provider.put_account('RubySMB', 'password')  # password can also be an NTLM hash

server = RubySMB::Server.new(
  gss_provider: ntlm_provider
)
puts "server is running"
server.run do
  puts "received connection"
  true
end
```

## Administrivia
All of the important bits should be documented. Most of the code has some kind of basic unit test. What's notably missing from the unit tests are:

* a successful authentication transaction using the relevant GSS messages
* processing SMB requests and testing the expected state changes

## Testing
Using Windows is one of the most comprehensive ways to test the authentication process. Just tap `Win + R` and then enter the UNC path into the server where the test harness is running, e.g. `\\1.2.3.4\share\somefile.txt`. It can be a bit slow at times while Windows is negotiating the connection. Delays of up to a minute were observed.

For each of the major SMB versions 1, 2 and 3:
* Checkout [this documentation](https://docs.microsoft.com/en-us/windows-server/storage/file-server/troubleshoot/detect-enable-and-disable-smbv1-v2-v3) for steps to enable and disable different SMB versions. Make sure that the SMB version of the **client** is changed (not the server) since the client is what is being used to run the tests.
  - [x] Insert bogus credentials and see that authentication fails
  - [x] Insert correct credentials and see authentication succedds
  - [x] See an incoming `TreeConnect` request

For at least one version:
- [x] Test a Net-NTLMv1 authentication process
    * Easiest way is to use Windows 7 and see [this documentation](https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/network-security-lan-manager-authentication-level) on setting the authentication level via Group Policy
For at least one version:
- [x] Test that anonymous authentication can be completed
    * This was easiest to test using the Nautilus file manager on Linux

If you use the `example/authenticate.rb` script, there **will not** be a `TreeConnect` request after the authentication is complete. That only occurs if the client is attempting to access a file.

If any issues are encountered, please collect a PCAP for debugging purposes and note the expected dialect from the testing environment.


examples/auth_capture.rb
```diff
@@ -0,0 +1,70 @@
...
+      when :ntlmv1
+        hash_type = 'NTLMv1-SSP'
+        hash << ":#{bin_to_hex(type3_msg.lm_response)}"
+        hash << ":#{bin_to_hex(type3_msg.ntlm_response)}"
+        hash << ":#{bin_to_hex(@server_challenge)}"
+      when :ntlmv2
+        hash_type = 'NTLMv2-SSP'
+        hash << ":#{bin_to_hex(@server_challenge)}"
+        hash << ":#{bin_to_hex(type3_msg.ntlm_response[0...16])}"
+        hash << ":#{bin_to_hex(type3_msg.ntlm_response[16.. -1])}"
```
agalway-r7, line 35: Could put a comment in here explaining why NTLMv2 pulls both parts of it's hash from the one variable?

lib/ruby_smb/client/signing.rb
```diff
@@ -1,64 +0,0 @@
-module RubySMB
```
agalway-r7, line 1: Why did you pull the signing module out of the client?

lib/ruby_smb/gss.rb
```diff
@@ -27,76 +48,54 @@ def self.asn1encode(str = '')
     # Create a GSS Security Blob of an NTLM Type 1 Message.
     # This code has been cargo culted and needs to be researched
     # and refactored into something better later.
-    # @todo Refactor this into non-magical code
```
agalway-r7, line 30: lol

lib/ruby_smb/gss.rb
```diff
@@ -27,76 +48,54 @@ def self.asn1encode(str = '')
     # Create a GSS Security Blob of an NTLM Type 1 Message.
     # This code has been cargo culted and needs to be researched
     # and refactored into something better later.
```
agalway-r7, line 50: Remove these lines if you have improved the code?

lib/ruby_smb/gss.rb
```diff
@@ -27,76 +48,54 @@ def self.asn1encode(str = '')
...
+              OpenSSL::ASN1::OctetString.new(type1)
+            ], 2, :CONTEXT_SPECIFIC)
+          ])
+        ], 0, :CONTEXT_SPECIFIC)
+      ], 0, :APPLICATION).to_der
     end
 
     # Create a GSS Security Blob of an NTLM Type 2 Message.
     # This code has been cargo culted and needs to be researched
     # and refactored into something better later.
```
agalway-r7, line 71: as above

lib/ruby_smb/gss/provider.rb
```diff
@@ -0,0 +1,35 @@
...
+        # Whether or not anonymous authentication attempts should be permitted.
+        #
+        attr_accessor :allow_anonymous
+      end
+    end
+  end
+end
+
+require 'ruby_smb/gss/provider/authenticator'
+require 'ruby_smb/gss/provider/ntlm'
```
agalway-r7, line 35: This gives me anxiety 😅  Can we put them on the top of the file?

lib/ruby_smb/gss/provider/authenticator.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Gss
+    module Provider
+      module Authenticator
+        #
+        # The base class for a GSS provider's unique authenticator. This provides a common interface and is not usable
+        # on it's own. The provider-specific authentication logic is defined within this authenticator class which
+        # actually runs the authentication routine.
+        #
+        class Base
+          # @param [Provider::Base] provider the GSS provider that this instance is an authenticator for
```
agalway-r7, line 11:
```suggestion
          # @param [Provider::Base] the GSS provider that this instance is an authenticator for
```

lib/ruby_smb/gss/provider/authenticator.rb
```diff
@@ -0,0 +1,42 @@
...
+          # @param server_client the client instance that this the authenticator will be for
```
agalway-r7, line 12:
```suggestion
          # @param server_client the client instance that this will be an authenticator for
```

lib/ruby_smb/gss/provider/authenticator.rb
```diff
@@ -0,0 +1,42 @@
...
+            @server_client = server_client
+            @session_key = nil
+            reset!
+          end
+
+          #
+          # Process a GSS authentication buffer. If no buffer is specified, the request is assumed to be the first in
+          # the negotiation sequence.
+          #
+          # @param [String, nil] buffer the request GSS request buffer that should be processed
```
agalway-r7, line 24:
```suggestion
          # @param [String, nil] the GSS request buffer that should be processed
```

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+                  OpenSSL::ASN1::Sequence.new([
+                    OpenSSL::ASN1::ASN1Data.new([
+                      OpenSSL::ASN1::Sequence.new([
+                        Gss::OID_NTLMSSP
+                      ])
+                    ], 0, :CONTEXT_SPECIFIC),
+                    OpenSSL::ASN1::ASN1Data.new([
+                      OpenSSL::ASN1::ASN1Data.new([
+                        OpenSSL::ASN1::ASN1Data.new([
+                          OpenSSL::ASN1::GeneralString.new('not_defined_in_RFC4178@please_ignore')
```
agalway-r7, line 77: I get it's a niche edgecase, but would it be worth changing this to something that an IDS will be less likely to pick up on. Could be mistaken but I'm sure I saw `not_defined_in_RFC4178@please_ignore` in a packet and it stuck out like a sore thumb.

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+              result = process_gss_type3(gss_api)
+            end
+
+            result
+          end
+
+          #
+          # Process the NTLM type 1 message and build a type 2 response message.
+          #
+          # @param [Net::NTLM::Message::Type1] type1_msg the NTLM type 1 message received by the client that should be
```
agalway-r7, line 105:
```suggestion
          # @param [Net::NTLM::Message::Type1] the NTLM type 1 message received by the client that should be
```

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+          def process_ntlm_type1(type1_msg)
+            type2_msg = Net::NTLM::Message::Type2.new.tap do |msg|
+              msg.target_name = 'LOCALHOST'.encode('UTF-16LE').b
+              msg.flag = 0
+              %i{ KEY56 KEY128 KEY_EXCHANGE UNICODE TARGET_INFO VERSION_INFO }.each do |flag|
+                msg.flag |= NEGOTIATE_FLAGS.fetch(flag)
+              end
+
+              @server_challenge = @provider.generate_server_challenge
+              msg.challenge = @server_challenge.unpack1('Q')
```
agalway-r7, line 117: Comments on pack always helps with the eyebleed, something like:
`Q => 64-bit unsigned, native endian (uint64_t)`

lib/ruby_smb/gss/provider/ntlm.rb
zeroSteiner: That's actually from the MS docs, see [[MS-SPNG] section 2.2.1 NegTokenInit2:](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-spng/8e71cf53-e867-4b79-b5b5-38c92be3d472)

> hintName: SHOULD<5> contain the string "not_defined_in_RFC4178@please_ignore".

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+            Result.new(buffer, nt_status, identity)
+          end
+        end
+
+        # @param [Boolean] allow_anonymous whether or not to allow anonymous authentication attempts
+        # @param [String] default_domain the default domain to use for authentication, unless specified 'WORKGROUP' will
+        #   be used
+        def initialize(allow_anonymous: false, default_domain: nil)
+          @allow_anonymous = allow_anonymous
+          @default_domain = default_domain || 'WORKGROUP'
```
agalway-r7, line 275:
```suggestion
        def initialize(allow_anonymous: false, default_domain: 'WORKGROUP')
          @allow_anonymous = allow_anonymous
          @default_domain = default_domain
```

Would this do the same thing?

zeroSteiner: Doing it like `def initialize(allow_anonymous: false, default_domain: 'WORKGROUP')` would allow the `default_domain` to be nil if someone explicitly set it to nil which would cause errors later on. I was trying to avoid that since a default domain is necessary. I guess I could raise an arg error if the domain is invalid. That might be more clear.

lib/ruby_smb/server.rb
```diff
@@ -0,0 +1,39 @@
...
+    require 'ruby_smb/gss/provider/ntlm'
+
+    Connection = Struct.new(:client, :thread)
+
+    # @param server_sock the socket on which the server should listen
+    # @param [Gss::Provider] the authentication provider
+    def initialize(server_sock: nil, gss_provider: nil)
+      server_sock = ::TCPServer.new(445) if server_sock.nil?
+
+      @server_guid = Random.new.bytes(16)
```
agalway-r7, line 18: Why is this `Random.new.bytes` when the NTLM generates challenges with  `SecureRandom.bytes`?

lib/ruby_smb/server/server_client.rb
```diff
@@ -0,0 +1,163 @@
...
+      # authenticated.
+      #
+      # @param [String] raw_request the request that should be handled
+      def handle_authenticated(raw_request)
+        response = nil
+
+        case raw_request[0...4]
+        when "\xff\x53\x4d\x42".b
+          raise NotImplementedError
+        when "\xfe\x53\x4d\x42".b
```
agalway-r7, line 65: Human readable comment for what these represent would be nice 👍

lib/ruby_smb/server.rb
zeroSteiner: The Server GUID isn't sensitive. It can be sent to unauthenticated entities so it didn't seem worthwhile to take it from the potentially limited source of secure entropy.

lib/ruby_smb/server/server_client/negotiation.rb
```diff
@@ -0,0 +1,149 @@
...
+
+          nil
+        end
+
+        def do_negotiate_smb1(request)
+          if request.dialects.map(&:dialect_string).include?(Client::SMB1_DIALECT_SMB2_WILDCARD)
+            response = SMB2::Packet::NegotiateResponse.new
+            response.smb2_header.credits = 1
+            response.security_mode.signing_enabled = 1
+            response.dialect_revision = 0x02ff
```
agalway-r7, line 37: Is there a human readable version of dialect `0x02ff`?

lib/ruby_smb/server/server_client/session_setup.rb
```diff
@@ -0,0 +1,82 @@
...
+          nil
+        end
+
+        def do_session_setup_smb1(request)
+          gss_result = process_gss(request.data_block.security_blob)
+          return if gss_result.nil?
+
+          response = SMB1::Packet::SessionSetupResponse.new
+          response.smb_header.pid_low = request.smb_header.pid_low
+          response.smb_header.uid = rand(0x10000)
```
agalway-r7, line 37: Is this `rand` to cut down on computation time by not calling `SecureRandom` all the time?

lib/ruby_smb/gss/provider.rb
zeroSteiner: I don't think they can be. When I place them up at the top of the file, I receive the following error:

```
  !: ruby_smb:feat/server12:46:14 ruby_smb git diff
diff --git a/lib/ruby_smb/gss/provider.rb b/lib/ruby_smb/gss/provider.rb
index b68be39..97a1241 100644
--- a/lib/ruby_smb/gss/provider.rb
+++ b/lib/ruby_smb/gss/provider.rb
@@ -1,3 +1,6 @@
+require 'ruby_smb/gss/provider/authenticator'
+require 'ruby_smb/gss/provider/ntlm'
+
 module RubySMB
   module Gss
     #
@@ -30,6 +33,3 @@ module RubySMB
     end
   end
 end
-
-require 'ruby_smb/gss/provider/authenticator'
-require 'ruby_smb/gss/provider/ntlm'
  !: ruby_smb:feat/server12:46:16 ruby_smb ruby examples/auth_capture.rb 
Traceback (most recent call last):
	14: from examples/auth_capture.rb:4:in `<main>'
	13: from examples/auth_capture.rb:4:in `require'
	12: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb.rb:13:in `<top (required)>'
	11: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb.rb:17:in `<module:RubySMB>'
	10: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb.rb:17:in `require'
	 9: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss.rb:1:in `<top (required)>'
	 8: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss.rb:4:in `<module:RubySMB>'
	 7: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss.rb:5:in `<module:Gss>'
	 6: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss.rb:5:in `require'
	 5: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss/provider.rb:2:in `<top (required)>'
	 4: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss/provider.rb:2:in `require'
	 3: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss/provider/ntlm.rb:1:in `<top (required)>'
	 2: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss/provider/ntlm.rb:2:in `<module:RubySMB>'
	 1: from /home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss/provider/ntlm.rb:3:in `<module:Gss>'
/home/smcintyre/Repositories/ruby_smb/lib/ruby_smb/gss/provider/ntlm.rb:8:in `<module:Provider>': uninitialized constant RubySMB::Gss::Provider::Base (NameError)
Did you mean?  Base64
exit status: 1                                                                                                                                                                                                                                                                                                              
  !: ruby_smb:feat/server12:46:22 ruby_smb 
```

Which I'm pretty confident is related to the fact that the `RubySMB::Gss::Provder` module needs to be defined before those files are loaded. There's a similar pattern in the existing [smb1.rb](https://github.com/rapid7/ruby_smb/blob/53e4855b76a63f9ded497182b07a9181b820a925/lib/ruby_smb/smb1.rb#L8) and [smb2.rb](https://github.com/rapid7/ruby_smb/blob/53e4855b76a63f9ded497182b07a9181b820a925/lib/ruby_smb/smb2.rb#L9). In each of those however the requires are within the module definition. I can move them there if that would be better, but I don't think they can be at the top.

spec/lib/ruby_smb/client_spec.rb
```diff
@@ -223,15 +225,11 @@
...
+          allow(smb2_client).to receive(:session_key).and_return(Random.new.bytes(16))
           expect(smb2_client).to receive(:smb2_sign).with(smb2_request).and_call_original
           smb2_client.send_recv(smb2_request)
         end
-
-        it 'calls #smb3_sign if it is an SMB3 client' do
-          allow(smb3_client).to receive(:is_status_pending?).and_return(false)
-          expect(smb3_client).to receive(:smb3_sign).with(smb2_request).and_call_original
-          smb3_client.send_recv(smb2_request)
-        end
```
agalway-r7, line 234: Are we testing sm3_sign elsewhere?

lib/ruby_smb/gss/provider.rb
agalway-r7: No that makes sense! Thanks for the explanation 👍

lib/ruby_smb/gss/provider/ntlm.rb
agalway-r7: Kewl, if it's the default value then it no worries

lib/ruby_smb/gss/provider/ntlm.rb
agalway-r7: Yeah an arg error would make it clearer, especially if someone is just jumping into the method definition to quickly check what is and isn't allowed

spec/lib/ruby_smb/client_spec.rb
zeroSteiner: That broke because the original `smb3_sign` function would check to see if the packet should be signed or not and only sign it when appropriate. Since I moved that into a mixin so it could be used by both, I switched the logic to **always** sign and made it the caller's responsibility to determine whether or not the signing operation should be performed. This was necessary because the check is different for the client and server. I also thought it was a little misleading that the function named `smb3_sign` may or may not actually sign the data 🤷🏻 .

Commit for context: 8998b4856fab1ea009282cb1f0d8eb204e58ce08

FWIW, the original test was calling the `smb3_sign` function, but the signing wasn't actually being performed so I think the value it was providing was negligible.

lib/ruby_smb/client/signing.rb
zeroSteiner: So it could be used by both the client and the server.

spec/lib/ruby_smb/client_spec.rb
agalway-r7: Agreed, especially with the `smb3_sign` only sometimes actually signing

cdelafuente-r7 (commented):
Thanks @zeroSteiner for adding this! I reviewed the code (except the specs) and left some comments for improvements. Otherwise it looks very good to me and it works perfectly. I still need to review the specs and finish testing with different clients. I'll let you know if something comes out.

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+          uint16 :build
+          uint32 :ntlm_revision, initial_value: 15
+
+          def to_s
+            "Version #{major}.#{minor} (Build #{build}); NTLM Current Revision #{ntlm_revision}"
+          end
+        end
+
+        class Authenticator < Authenticator::Base
+          NEGOTIATE_FLAGS = {
```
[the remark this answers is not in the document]
zeroSteiner: I actually was using the `Net::NTLM` flag definitions when I first started by found them difficult to use while debugging for two reasons.
1. Using the hex values made it difficult to understand what bit field each corresponded to.
2. As you noticed, the flag names are different. I was using Wireshark to debug some issues and some of the names were different and others were missing. This made for a cumbersome experience. Take `MBZ7` for example, I have no idea what that is but it corresponds to Anonymous in Wireshark.

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+                  ])
+                ], 0, :CONTEXT_SPECIFIC)
+              ], 0, :APPLICATION).to_der
+              return Result.new(buffer, WindowsError::NTStatus::STATUS_SUCCESS)
+            end
+
+            begin
+              gss_api = OpenSSL::ASN1.decode(request_buffer)
+            rescue OpenSSL::ASN1::ASN1Error
+              return
```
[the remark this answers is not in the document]
zeroSteiner: Yup, I looked around a couple of times to see if we had some kind of logging capability that I could use because you're right. There's a few instances where some kinda error like this will result in the connection being closed.

lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -0,0 +1,337 @@
...
+        #
+        # @param [String] username the username of the account to fetch.
+        # @param [String, nil] domain the domain in which the account to fetch exists.
+        # @return [Account, nil] the account if it was found
+        def get_account(username, domain: nil)
+          # the username and password values should use the native encoding for the comparison in the #find operation
+          username = username.downcase
+          domain = @default_domain if domain.nil? || domain == '.'.encode(domain.encoding)
+          domain = domain.downcase
+          @accounts.find { |account| account.username.encode(username.encoding).downcase == username && account.domain.encode(domain.encoding).downcase == domain }
```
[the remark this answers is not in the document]
zeroSteiner: Wouldn't the computer name be the authentication domain though? Say you have two non-domain joined computers both with users "bob" on them. Those are two separate users with two separate passwords that are valid in two separate contexts. It's quite feasible that bob from system A is a user we want to authenticate while bob from system B is not.

I think in this case, it's working as intended right? That's why when the user wants to authenticate, they should specify the `WORKGROUP` or `.` domain because their local computer account probably isn't valid.

If we really wanted though, maybe an option to just ignore the domain could be added to the provider.

lib/ruby_smb/server.rb
```diff
@@ -0,0 +1,39 @@
...
+  # available at this time. The negotiating and authentication is supported for SMB versions 1 through 3.1.1.
+  class Server
+    require 'ruby_smb/server/server_client'
+    require 'ruby_smb/gss/provider/ntlm'
+
+    Connection = Struct.new(:client, :thread)
+
+    # @param server_sock the socket on which the server should listen
+    # @param [Gss::Provider] the authentication provider
+    def initialize(server_sock: nil, gss_provider: nil)
```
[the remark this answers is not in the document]
zeroSteiner: But the `RubySMB::Dispatcher` doesn't have a method to accept client connections, does it? It was my understanding that it wasn't intended to wrap server/listening sockets but rather established connections. Also a read timeout on a listening socket I'm pretty sure corresponds to a connection timeout as in how long to wait for a client to connect.

lib/ruby_smb/server.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    # Run the server and accept any connections. For each connection, the block will be executed if specified. When the
+    # block returns false, the loop will exit and the server will no long accept new connections.
+    def run(&block)
+      loop do
+        sock = @server_sock.accept
+        server_client = ServerClient.new(self, RubySMB::Dispatcher::Socket.new(sock))
+        @connections << Connection.new(server_client, Thread.new { server_client.run })
+
+        break unless block.nil? || block.call(server_client)
```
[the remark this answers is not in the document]
zeroSteiner: I'm not sure what you mean, can you provide an example? The loop body needs to be executed either way to accept a connection and initialize a new `ServerClient` instance. The `block.nil?` check is only present to allow the block to be optional.

lib/ruby_smb/server/server_client/session_setup.rb
```diff
@@ -0,0 +1,82 @@
...
+          response.smb2_header.session_id = @session_id = @session_id || SecureRandom.random_bytes(4).unpack1('V')
+          response.buffer = gss_result.buffer
+
+          update_preauth_hash(request) if @dialect == '0x0311'
+          if gss_result.nt_status == WindowsError::NTStatus::STATUS_SUCCESS
+            @state = :authenticated
+            @identity = gss_result.identity
+            @session_key = @gss_authenticator.session_key
+          elsif @dialect == '0x0311'
+            update_preauth_hash(response)
```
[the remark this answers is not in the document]
zeroSteiner: I'll update it to only update the hash when the dialect is 3.1.1 **and** the status is `STATUS_MORE_PROCESSING_REQUIRED`.

lib/ruby_smb/smb2/packet/session_setup_response.rb
```diff
@@ -11,8 +11,8 @@ class SessionSetupResponse < RubySMB::GenericPacket
         uint16              :structure_size, label: 'Structure Size', initial_value: 9
         session_flags       :session_flags
         uint16              :security_buffer_offset,  label: 'Security Buffer Offset', initial_value: 0x48
-        uint16              :security_buffer_length,  label: 'Security Buffer Length'
-        string              :buffer,                  label: 'Security Buffer', length: -> { security_buffer_length }
+        uint16              :security_buffer_length,  label: 'Security Buffer Length', value: -> { buffer.length }
```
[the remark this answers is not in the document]
zeroSteiner: That makes a lot of sense, thanks for explaining the difference!

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
cdelafuente-r7: Absolutely, there are a few undocumented/unknown flags in the `Net::NTLM` library, including the Anonymous flag. I fixed them in the [PR](https://github.com/WinRb/rubyntlm/pull/38/files#diff-d0a5655805436a73145e2fed15c8916c907f9874e89670269615d015041c54bcR16) I submitted. It is based on this [documentation](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/99d90ff4-957f-4c8a-80e4-5bfe5a9a9832?redirectedfrom=MSDN).

That said, I agree that the bit field matches the documentation, so we can use your implementation. Would it be possible to move them in a more generic location and update the client to use them instead of `Net::NTLM` flags? I believe there is just one place where these flags are used: https://github.com/rapid7/ruby_smb/blob/master/lib/ruby_smb/client.rb#L652-L660

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
cdelafuente-r7: If I understand your example correctly, the two users belong to two separate hosts, since they are local users (non-domain users). So, there will be two separate SMB servers in this case. Each one will authenticate the user locally in its own context. I think that any authentication of a user without domain is considered local, and the domain should be ignored.

I tested this with the `net use` command:
```
C:\Users\TestUserDA>net use \\172.31.41.227\share /user:foo\administrator
The password or user name is invalid for \\172.31.41.227\share.

Enter the password for 'foo\administrator' to connect to '172.31.41.227':
The command completed successfully.
```

It successfully logged on 172.31.41.227, which is on a different domain, with the local administrator. The `NTLNSSP` blob in the `Session Setup` request contains `foo` as domain name:

Screenshot 2021-08-18 at 21 13 13

It looks like the domain name is ignored.

That said, I'm not sure if this applies in the context of this SMB server.

lib/ruby_smb/server.rb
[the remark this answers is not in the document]
cdelafuente-r7: You are absolutely right, I misread it... and need more coffee.

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
zeroSteiner: Fixed in commit c3894902521bdd29240dc5a60bf220ea209d3583. I also moved the `OSVersion` struct since that's a field in some of the NTLM messages and could be used elsewhere as well.

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
zeroSteiner: > So, there will be two separate SMB servers in this case. Each one will authenticate the user locally in its own context. 

There doesn't need to be two separate servers though. If I add accounts to my one server instance with two different domains:

```
ntlm_provider = RubySMB::Gss::Provider::NTLM.new
ntlm_provider.put_account('bob', 'password1', domain: 'HOST1')
ntlm_provider.put_account('bob', 'password2', domain: 'HOST2')
```

Then bob from HOST1 and bob from HOST2 will both be able to authenticate to me given that their systems will set the domain field to their unique hostnames. This then allows me to know the identity of each and differentiate HOST1\bob from HOST2\bob.

If we ignored the domain in the lookup, HOST1\bob and HOST2\bob would need to be the same account and share the same password which seems unnecessary. If we're mostly expecting HOST1\bob to connect to us, then we could set the default domain to HOST1, and that would just mean that HOST2\bob would need to specify their hostname as the domain field which you mentioned was the behavior anyways.

lib/ruby_smb/server/server_client/negotiation.rb
```diff
@@ -0,0 +1,149 @@
...
+            response.server_guid = @server.server_guid
+
+            response.max_transact_size = 0x800000
+            response.max_read_size = 0x800000
+            response.max_write_size = 0x800000
```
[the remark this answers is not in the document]
zeroSteiner: Probably not right now because this is the only place where they're used. Once they need to be used elsewhere, then it would probably make sense to put them in constants or even as an instance variable if any of the values need to be client specific. Another option could be to set them using the `initial_value` of the `SMB2::Packet::NegotiateResponse` class if they're going to be server-specific and not unique to particular client connections.

lib/ruby_smb/server/server_client/negotiation.rb
```diff
@@ -0,0 +1,149 @@
...
+          response.data_block.server_guid = @server.server_guid
+          response.data_block.security_blob = process_gss.buffer
+
+          @state = :session_setup
+          @dialect = dialect
+          response
+        end
+
+        def do_negotiate_smb2(request)
+          dialect = ([0x311, 0x302, 0x300, 0x210, 0x202] & request.dialects.map(&:to_i)).first
```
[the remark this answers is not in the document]
zeroSteiner: Added in commit 0f884d617a1e962b0e4af1955e64e8ac03f24bcf. You can now configure the supported dialects and their preferences by doing something like `server.dialects.reject! { |dialect| RubySMB::Dialect[dialect].family == RubySMB::Dialect::FAMILY_SMB3 }` to disable SMB 3.

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
cdelafuente-r7: Absolutely, it makes sense this way. I didn't think of using the server to host the same local account name from different hosts, which requires the hostname to differentiate them. We will just need to know the correct hostname in advance, which I believe is is not a big deal.

lib/ruby_smb/server.rb
```diff
@@ -0,0 +1,54 @@
...
+    attr_reader :dialects
+
+    # The GSS Provider instance that this server will use to authenticate
+    # incoming client connections.
+    # @!attribute [r] gss_provider
+    #   @return [RubySMB::Gss::Provider::Base]
+    attr_reader :gss_provider
+
+    # The 16 byte GUID that uniquely identifies this server instance.
+    # @!attribute [r] server_guid
```
cdelafuente-r7, line 50:
```suggestion
    # @!attribute [r] guid
```

lib/ruby_smb/server.rb
```diff
@@ -0,0 +1,54 @@
...
+        @connections << Connection.new(server_client, Thread.new { server_client.run })
+
+        break unless block.nil? || block.call(server_client)
+      end
+    end
+
+    # The dialects that this server will negotiate with clients, in ascending order of preference.
+    # @!attribute [r] dialects
+    #   @return [Array<String>]
+    attr_reader :dialects
```
cdelafuente-r7, line 41: Should these attributes be writable? This would let the user choose the dialects. I believe this also might be the case for the other attributes.

zeroSteiner: It doesn't *need* to be because the array container is a mutable object. It probably makes sense to be writable though so I'll update it.

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
sempervictus: Since this is effectively analogous to Rex::Proto::SMB, can we implement logging a la Rex?

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
sempervictus: I think we want to leave this exposed to API and even user configuration as much as possible since the semantics of protocol access patterns inform defensive analysis - the more entropy we feed into whatever we put on the wire, the less chance of a fingerprint or heuristic match being built around our interactions with target systems.

lib/ruby_smb/server/server_client/negotiation.rb
```diff
@@ -0,0 +1,149 @@
...
+            response.data_block.byte_count = 0
+            return response
+          end
+
+          response = SMB1::Packet::NegotiateResponseExtended.new
+          response.parameter_block.dialect_index = dialect_strings.index(dialect)
+          response.parameter_block.max_mpx_count = 50
+          response.parameter_block.max_number_vcs = 1
+          response.parameter_block.max_buffer_size = 16644
+          response.parameter_block.max_raw_size = 65536
```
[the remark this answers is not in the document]
sempervictus: +1 - reading this is a bit painful :).

lib/ruby_smb/signing.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      packet.smb2_header.flags.signed = 1
+      packet.smb2_header.signature = "\x00" * 16
+      hmac = OpenSSL::CMAC.digest('AES', signing_key, packet.to_binary_s)
+      packet.smb2_header.signature = hmac[0, 16]
+
+      packet
+    end
+  end
+end
```
sempervictus, line 59: Do we want to have the sig-stripping and MITM-related functions in here too?

sempervictus:
Did a first pass - awesome stuff, thanks folks

lib/ruby_smb/gss/provider/ntlm.rb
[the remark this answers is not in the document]
zeroSteiner: Yeah I think that's the way to go. I'll see about adding that on the next major dev pass.

lib/ruby_smb/signing.rb
cdelafuente-r7: That's a very suggestion. I do think we will need these capabilities in this SMB server at some point. But, right now, we're not ready to use them yet, so we don't know exactly the details. I think we should focus on basic functionalities for this first implementation and go back to more advanced features once we know exactly what is needed.

We'll definitely keep this in mind for future iterations. Thanks for the heads up!

lib/ruby_smb/server/server_client/negotiation.rb
```diff
@@ -0,0 +1,153 @@
...
+            disconnect!
+          else
+            send_packet(response)
+          end
+
+          nil
+        end
+
+        def do_negotiate_smb1(request)
+          if request.dialects.map(&:dialect_string).include?(Client::SMB1_DIALECT_SMB2_WILDCARD)
```
cdelafuente-r7, line 33: I found an issue while testing dialect selection with SMBv1. I used this script to test:
```ruby
#!/usr/bin/ruby

require 'bundler/setup'
require 'ruby_smb'
require 'ruby_smb/gss/provider/ntlm'

ntlm_provider = RubySMB::Gss::Provider::NTLM.new(allow_anonymous: true, default_domain: '')
ntlm_provider.put_account('RubySMB', 'password')

server = RubySMB::Server.new(
  gss_provider: ntlm_provider
)
server.dialects = ['NT LM 0.12']

puts "server is running"
server.run do
  puts "received connection"
  true
end
```

The issue is that, most of the time, the client sends a `SMB1_DIALECT_SMB2_WILDCARD` first. So, if the server is setup to only accept SMBv1, like in the example above, it will never select this protocol version and always return a SMBv2 response with `SMB2::SMB2_WILDCARD_REVISION`. It is not possible to force the client to use SMBv1.

I think we should check if the server supports any SMBv2 dialect before sending a wild card SMBv2 response. Something along these lines:
```suggestion
          smb2_dialects = @server.dialects.select { |dialect| Dialect[dialect].order == Dialect::ORDER_SMB2 }
          if request.dialects.map(&:dialect_string).include?(Client::SMB1_DIALECT_SMB2_WILDCARD) && !smb2_dialects.empty?
```

zeroSteiner: Great catch! You're right, I'll make that change. I don't think I tested only SMB1.

cdelafuente-r7:
# Release Notes
This adds an initial SMB server implementation. It supports SMBv1 (only `NT LM 0.12` dialect) through SMBv3 and every dialect in the 2.x and 3.x families. This server can perform authentication, including protocol negotiation and session setup.

PR #180, opened 2021-11-23
Use the correct message ID in SMB2/3 responses

zeroSteiner:
This fixes a message ID synchronization issue between the requests and the responses that would cause there to be a noticeable delay when a Windows client authenticates to the server.

To test this and reproduce the original issue run the `examples/auth_capture.rb` script. Use a Windows client that uses SMB2/3 (I tested with Server 2019) and connect to the capture server. Notice that it takes a while ~30 seconds or so. You can load Wireshark and follow the negotiation and session setup requests and replies and notice that the message ID is out of sync.

With this change in place, the correct message ID is used and it appears to fix the delay causing the connection to occur much faster. In the example of the auth capture server, the authentication/connection will still fail which is intended, however, it'll do so without the delay.

### DIFF
--- a/lib/ruby_smb/server/server_client.rb
+++ b/lib/ruby_smb/server/server_client.rb
@@ -22,7 +22,6 @@ module RubySMB
         @dispatcher = dispatcher
         @state = :negotiate
         @dialect = nil
-        @message_id = 0
         @session_id = nil
         @session_key = nil
         @gss_authenticator = server.gss_provider.new_authenticator(self)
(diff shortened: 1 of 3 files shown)


cdelafuente-r7:
Thanks @zeroSteiner for this fix! I confirm this fixes the delay issue. Also, I noticed the Windows client was retrying the negotiation/session setup sequence multiple times after the ~30 seconds delays. This also fixes this, now it is done just once, without delays.

One small thing I would like to check before landing: it looks like `@message_id` instance variable is not used anymore. If it is the case, should we keep it in the server client ([here](https://github.com/rapid7/ruby_smb/blob/b89636748f392f4fcf6b353ba0952923d386bf98/lib/ruby_smb/server/server_client.rb#L25))?

zeroSteiner:
That's right, `@message_id` isn't used anymore. At the time I thought I might need it in the server but I'm a lot further along now and more confident that's not the case so I think we can remove it. I'll push that change up now. Thanks for reviewing and testing!

PR #185, opened 2022-01-04
Bump the minor version

cdelafuente-r7:
Since the redesign of the NDR structures changed some exposed API's, bumping the patch version was not enough.


smcintyre-r7:
@jmartin-r7 Are we sure this should be the minor version and not the major version per [item 8](https://semver.org/#spec-item-8) in the Semantic Versioning spec?

jmartin-r7:
@smcintyre-r7, you are correct the exposed object are public and not backwards compatible so this would warrant a `major` bump.

PR #189, opened 2022-01-26
Changes to support smb_relay versions smb2/smb3

adfoster-r7:
Draft pull request with minimal changes to allow Metasploit-framework to support ntlm relaying with smb2/3

Framework PR https://github.com/rapid7/metasploit-framework/pull/16098

### DIFF
--- a/lib/ruby_smb/client.rb
+++ b/lib/ruby_smb/client.rb
@@ -29,10 +29,26 @@ module RubySMB
     # It indicates that the server implements SMB 2.1 or future dialect revisions
     # Note that this must be used for SMB3
     SMB1_DIALECT_SMB2_WILDCARD = 'SMB 2.???'.freeze
+
+    SMB2_DIALECT_0202 = '0x0202'.freeze
+    SMB2_DIALECT_0210 = '0x0210'.freeze
+    SMB2_DIALECT_0300  = '0x0300'.freeze
+    SMB2_DIALECT_0302  = '0x0302'.freeze
+    SMB2_DIALECT_0311  = '0x0311'.freeze
+
     # Dialect values for SMB2
-    SMB2_DIALECT_DEFAULT = ['0x0202', '0x0210']
+    SMB2_DIALECT_DEFAULT = [
+      SMB2_DIALECT_0202,
+      SMB2_DIALECT_0210,
+    ].freeze
+
     # Dialect values for SMB3
-    SMB3_DIALECT_DEFAULT = ['0x0300', '0x0302', '0x0311']
+    SMB3_DIALECT_DEFAULT = [
+      SMB2_DIALECT_0300,
+      SMB2_DIALECT_0302,
+      SMB2_DIALECT_0311
+    ].freeze
+
     # The default maximum size of a SMB message that the Client accepts (in bytes)
     MAX_BUFFER_SIZE = 64512
     # The default maximum size of a SMB message that the Server accepts (in bytes)
--- a/lib/ruby_smb/client/negotiation.rb
+++ b/lib/ruby_smb/client/negotiation.rb
@@ -251,7 +251,7 @@ module RubySMB
           raise ArgumentError, 'Must be an array of strings' unless dialect.is_a? String
           packet.add_dialect(dialect.to_i(16))
         end
-        packet.capabilities.encryption = 1
+        packet.capabilities.encryption = @session_encrypt_data ? 1 : 0
 
         if packet.dialects.include?(0x0311)
           nc = RubySMB::SMB2::NegotiateContext.new(
--- a/lib/ruby_smb/server/session.rb
+++ b/lib/ruby_smb/server/session.rb
@@ -13,6 +13,7 @@ module RubySMB
         @user_id = user_id
         @state = state
         @signing_required = false
+        @metadata = {}
         # tree id => provider processor instance
         @tree_connect_table = {}
         @creation_time = Time.now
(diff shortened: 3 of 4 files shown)


lib/ruby_smb/client.rb
```diff
@@ -444,7 +444,8 @@ def send_recv(packet, encrypt: false)
         unless packet.is_a?(RubySMB::SMB2::Packet::SessionSetupRequest) || session_key.empty?
           if self.smb2 && signing_required
             packet = smb2_sign(packet)
-          elsif self.smb3 && (signing_required || packet.is_a?(RubySMB::SMB2::Packet::TreeConnectRequest))
+          elsif self.smb3 && (signing_required)
+          # elsif self.smb3 && (signing_required || packet.is_a?(RubySMB::SMB2::Packet::TreeConnectRequest))
```
adfoster-r7, line 448: This will need extra work

lib/ruby_smb/smb2/packet/session_setup_request.rb
```diff
@@ -30,6 +30,17 @@ def set_type1_blob(type1_message)
...
+        # Takes a serialized NTLM Type 1 message inserts it into the {RubySMB::SMB2::Packet::SessionSetupRequest#buffer}
+        # as well as updating the {RubySMB::SMB2::Packet::SessionSetupRequest#security_buffer_length}.
+        # This method DOES NOT wrap the NTLM Type 1 message in GSS.
+        #
+        # @param type1_message [String] the serialized NTLM Type 1 message
+        # @return [void]
+        def set_ntlm_type1_blob(type1_message)
+          self.security_buffer_length = type1_message.length
+          self.buffer = type1_message
+        end
```
adfoster-r7, line 42: We'll probably want a better pattern here. Related thought, a potential rename of the newly added `gss_provider` namespace to something more generic may be beneficial as well

lib/ruby_smb/server/session.rb
```diff
@@ -62,6 +63,11 @@ def logoff!
       #   @return [Hash]
       attr_accessor :tree_connect_table
 
+      # Untyped hash for storing additional arbitrary metadata about the current session
+      # @!attribute [rw] metadaa
+      #   @return [Hash]
+      attr_accessor :metadata
```
adfoster-r7, line 69: I went with `metadata` at the time of implementation, but on reflection I'm thinking this is closer to `state` or just `data` in general

lib/ruby_smb/client/negotiation.rb
```diff
@@ -249,9 +249,11 @@ def smb2_3_negotiate_request
       def add_smb3_to_negotiate_request(packet, dialects = SMB3_DIALECT_DEFAULT)
         dialects.each do |dialect|
           raise ArgumentError, 'Must be an array of strings' unless dialect.is_a? String
+          next if dialect == '0x0311' && !@session_encrypt_data
```
smcintyre-r7, line 252: All of the SMB3 dialects support encryption.

adfoster-r7, line 252: Let's update ruby_smb's constructor to take in the dialects as configurable option instead of this approach

lib/ruby_smb/smb2/packet/session_setup_request.rb
```diff
@@ -30,6 +30,17 @@ def set_type1_blob(type1_message)
           self.buffer = gss_blob
         end
 
+        # Takes a serialized NTLM Type 1 message inserts it into the {RubySMB::SMB2::Packet::SessionSetupRequest#buffer}
+        # as well as updating the {RubySMB::SMB2::Packet::SessionSetupRequest#security_buffer_length}.
+        # This method DOES NOT wrap the NTLM Type 1 message in GSS.
+        #
+        # @param type1_message [String] the serialized NTLM Type 1 message
+        # @return [void]
+        def set_ntlm_type1_blob(type1_message)
```
smcintyre-r7, line 39: This looks like it's not necessary, the caller should be able to set the buffer directly. To remove the necessity on setting the length field correctly, you can update the field to use an `initial_value` lambda function.
```
        uint16                     :security_buffer_length,  label: 'Security Buffer Length', initial_value -> { buffer.length }
```

With this in place, you should be able to remove both the `set_ntlm_type1_blob` and `set_ntlm_type3_blob` functions. Alternatively, if we don't want to update that field, those two functions should probably be combined into a more generic one like `set_security_buffer` since the body of the functions are the same and they have nothing to do with NTLM.

lib/ruby_smb/client/negotiation.rb
adfoster-r7: Reworked, metasploit-framework now specifies the dialect versions required and this is now removed

lib/ruby_smb/smb2/packet/session_setup_request.rb
adfoster-r7: To keep things simple/flexible for the future, the previous methods have been consolidated into a single setter

lib/ruby_smb/client/negotiation.rb
adfoster-r7: This wasn't added to the constructor, instead it was overridden directly in framework

lib/ruby_smb/server/session.rb
adfoster-r7: This isn't a blocker for me, it turns out framework uses `context` - but I'm happy to stick with `metadata` for now

lib/ruby_smb/client.rb
adfoster-r7: Reverted entirely

smcintyre-r7 (commented):
Changes look good to me, I tested the psexec module in Metasploit using SMB 2 and 3, confirmed that 3 still automatically encrypts the connection and that both the PowerShell and Command targets are working. Tested the Command target because it performs extra steps to read a file.

Thanks @adfoster-r7 I'll get this landed in a moment.

smcintyre-r7 (approved):

PR #193, opened 2022-02-08
Fix File Read Operations

zeroSteiner:
Since commit 6cca27c, the open file operations no longer remove the `\` prefix on file paths. This broke file read operations as used by Metasploit. This resulted in execution failures when Metasploit's `psexec` module was used with the Command target which needs to read the file output. This readds the logic to remove the unnecessary prefix when opening files. This change has been tested using the newly updated read_file example on SMB versions 1, 2, 3 with and without a path with a `\` prefix.

- [ ] Use the example read_file to read a file path containing a leading `\`
- [ ] See the example script work
- [ ] Try it again with different SMB versions
    * Enable SMB 1 by following these instructions: https://docs.microsoft.com/en-us/windows-server/storage/file-server/troubleshoot/detect-enable-and-disable-smbv1-v2-v3

## Demo
This example shows reading a file from the temp directory that was created using psexec. SMB3 is used, but can be disabled using the CLI options on the example. Note that the file path contains a `\` prefix.
```
ruby read_file.rb --username smcintyre --password Password1 192.168.159.96 C$ "\\Windows\\Temp\\uDDauZxRI.txt"
SMB3 : (0x00000000) STATUS_WAIT_0: The caller specified WaitAny for WaitType and one of the dispatcher objects in the Object array has been set to the signaled state.
Connected to \\192.168.159.96\C$ successfully!

User accounts for \\

-------------------------------------------------------------------------------
$431000-ERJAFULIQUSS     Administrator            aliddle                  
Guest                    HealthMailbox0896988     HealthMailbox23c9938     
HealthMailbox256d6c9     HealthMailbox3dd648f     HealthMailbox3ef3d8d     
HealthMailbox8710d04     HealthMailbox8c45655     HealthMailboxa04271a     
HealthMailboxcad83a9     HealthMailboxde93ebe     HealthMailboxf4602f9     
jdoe                     krbtgt                   SM_2e9af84fe4c6473ea     
SM_3879f1f6de9b49818     SM_3c6a2d0a4c8648c9b     SM_5784a59a716948118     
SM_97642fc203314e6f8     SM_9db3f69793764fde9     SM_ac24b3f263bb427f9     
SM_c2d9a2c8d3844cd29     SM_f5faf24b3b1c46efa     smcintyre                
The command completed with one or more errors.

```

<details>
<summary>PSexec bug output</summary>

Using the "Command" target.

```
[*] 192.168.159.96:445 - Connecting to the server...
[*] 192.168.159.96:445 - Authenticating to 192.168.159.96:445 as user 'smcintyre'...
[+] 192.168.159.96:445 - Service start timed out, OK if running a command or non-service executable...
[-] 192.168.159.96:445 - Unable to get handle: The server responded with an unexpected status code: STATUS_INVALID_PARAMETER
[-] 192.168.159.96:445 - Command seems to still be executing. Try increasing RETRY and DELAY
[*] 192.168.159.96:445 - Getting the command output...
[-] 192.168.159.96:445 - Unable to read file \Windows\Temp\uDDauZxRI.txt. RubySMB::Error::UnexpectedStatusCode: The server responded with an unexpected status code: STATUS_INVALID_PARAMETER.
[-] 192.168.159.96:445 - Error getting command output
[*] 192.168.159.96:445 - Executing cleanup...
[-] 192.168.159.96:445 - Unable to cleanup \Windows\Temp\HvXgAIlbtIRw.bat. Error: The server responded with an unexpected status code: STATUS_INVALID_PARAMETER
[+] 192.168.159.96:445 - Cleanup was successful
[*] Exploit completed, but no session was created.
```
</details>

### DIFF
--- a/lib/ruby_smb/smb2/tree.rb
+++ b/lib/ruby_smb/smb2/tree.rb
@@ -60,13 +60,15 @@ module RubySMB
         # Make sure we don't modify the caller's hash options
         opts = opts.dup
         opts[:filename] = opts[:filename].dup
-        opts[:filename] = opts[:filename][1..-1] if opts[:filename].start_with? '\\'
+        opts[:filename] = opts[:filename][1..-1] if opts[:filename].start_with?('\\')
         open_file(**opts)
       end
 
       def open_file(filename:, attributes: nil, options: nil, disposition: RubySMB::Dispositions::FILE_OPEN,
                     impersonation: RubySMB::ImpersonationLevels::SEC_IMPERSONATE, read: true, write: false, delete: false)
 
+        filename = filename.dup
+        filename = filename[1..-1] if filename.start_with?('\\'.encode(filename.encoding))
         create_request = RubySMB::SMB2::Packet::CreateRequest.new
         create_request = set_header_fields(create_request)
 
--- a/lib/ruby_smb/smb2/tree.rb
+++ b/lib/ruby_smb/smb2/tree.rb
@@ -60,80 +60,16 @@ module RubySMB
         # Make sure we don't modify the caller's hash options
         opts = opts.dup
         opts[:filename] = opts[:filename].dup
-        opts[:filename] = opts[:filename][1..-1] if opts[:filename].start_with?('\\')
-        open_file(**opts)
+        opts[:filename] = opts[:filename][1..-1] if opts[:filename].start_with?('\\'.encode(opts[:filename].encoding))
+        _open(**opts)
       end
 
-      def open_file(filename:, attributes: nil, options: nil, disposition: RubySMB::Dispositions::FILE_OPEN,
-                    impersonation: RubySMB::ImpersonationLevels::SEC_IMPERSONATE, read: true, write: false, delete: false)
-
-        filename = filename.dup
-        filename = filename[1..-1] if filename.start_with?('\\'.encode(filename.encoding))
-        create_request = RubySMB::SMB2::Packet::CreateRequest.new
-        create_request = set_header_fields(create_request)
-
-        # If the user supplied file attributes, use those, otherwise set some
-        # sane defaults.
-        if attributes
-          create_request.file_attributes = attributes
-        else
-          create_request.file_attributes.directory  = 0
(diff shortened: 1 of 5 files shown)


zeroSteiner:
Oh yeah I see the issue you're referring to. What if instead of adding a new keyword argument though, I moved the bulk of the logic from `open_file` into a new function that both `open_pipe` and `open_file` called? Then `open_(file|pipe)` could update the filename parameter as appropriate before dispatching to the new function that would contain almost all the code currently in `open_file` that just wouldn't modify the filename parameter? Maybe a private `_open` method?

cdelafuente-r7:
Sounds perfect to me! go for it, please! Also, would it be possible to do it for both SMBv1 and SMBv2? I know there is no issue with SMBv2 and backslashes, but just for consistency.

lib/ruby_smb/smb2/tree.rb
```diff
@@ -60,78 +60,16 @@ def open_pipe(opts)
         # Make sure we don't modify the caller's hash options
         opts = opts.dup
         opts[:filename] = opts[:filename].dup
-        opts[:filename] = opts[:filename][1..-1] if opts[:filename].start_with? '\\'
+        opts[:filename] = opts[:filename][1..-1] if opts[:filename].start_with?('\\'.encode(opts[:filename].encoding))
         open_file(**opts)
```
cdelafuente-r7, line 64: I'm wondering if this should also be a call to `_open(**opts)` instead?

spec/lib/ruby_smb/smb1/tree_spec.rb
```diff
@@ -495,17 +495,17 @@
   describe '#open_pipe' do
```
cdelafuente-r7, line 495: That would be great to have specs that make sure the backslash is added or removed, according to wether `open_pipe` or `open_file` is called. This also applies for SMBv2.

lib/ruby_smb/smb2/tree.rb
smcintyre-r7: You're right I missed that. It's fixed now though.

PR #204, opened 2022-04-08
Add AES-256 support

### LINKED ISSUE #203: Add AES-256 Support
Looks like Microsoft recently added AES-256-CCM and AES-256-GCM encryption capabilities. Right now we only support AES-128. Hopefully it should be any easy thing to add. Difficulty will depend on if the key calculation needs to be updated and if our crypt libs support the algorithm. I remember having some trouble with one of the variant, so it's possible that the 256 version may not be available out of the box but I don't remember those details.

See:
* https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/16693be7-2b27-4d3b-804b-f605bde5bcdd
* https://twitter.com/NerdPyle/status/1511955762993524738

zeroSteiner:
Closes #203.

This adds AES-256 support which was recently added by Microsoft. Luckily the KDF remains the same. It also swaps the order of the algorithms, it looks like Windows expects them to be in order of priority. This will prefer the AES-256 variants, and fail back to the AES-128 variants. Once landed, Metasploit's smb_version module will automatically identify when AES-256 can be negotiated.

You need to test this with a fully up to date Windows 11 system. I don't think AES-256 is available for Windows 10. I also double checked that AES-128 negotiation and encryption still works correctly.

Server 2019 uses AES-128, while Windows 11 negotiates AES-256. This can be confirmed using Wireshark and inspecting the negotiate response frame form the server.
```
  : ruby_smb:feat/aes-25610:29:08 ruby_smb ruby examples/read_file.rb --username smcintyre --password "$(getpass)" 192.168.159.70 SMBShare readme.txt
Password: SMB3 : (0x00000000) STATUS_SUCCESS: The operation completed successfully.
Connected to \\192.168.159.70\SMBShare successfully!
Hello from Windows 11!
  : ruby_smb:feat/aes-25610:29:13 ruby_smb ruby examples/read_file.rb --username smcintyre --password "$(getpass)" 192.168.159.96 SMBShare readme.txt
Password: SMB3 : (0x00000000) STATUS_SUCCESS: The operation completed successfully.
Connected to \\192.168.159.96\SMBShare successfully!
Hello from Server 2019!
  : ruby_smb:feat/aes-25610:29:21 rub
```

This may introduce conflicts with 197 that will also need to be updated so the server also supports AES-256.

### DIFF
--- a/lib/ruby_smb/client/encryption.rb
+++ b/lib/ruby_smb/client/encryption.rb
@@ -4,6 +4,8 @@ module RubySMB
     module Encryption
       def smb3_encrypt(data)
         unless @client_encryption_key
+          raise RubySMB::Error::EncryptionError.new('The encryption algorithm has not be set') if @encryption_algorithm.nil?
+
           key_bit_len = OpenSSL::Cipher.new(@encryption_algorithm).key_len * 8
 
           case @dialect
@@ -37,6 +39,8 @@ module RubySMB
 
       def smb3_decrypt(th)
         unless @server_encryption_key
+          raise RubySMB::Error::EncryptionError.new('The encryption algorithm has not be set') if @encryption_algorithm.nil?
+
           key_bit_len = OpenSSL::Cipher.new(@encryption_algorithm).key_len * 8
 
           case @dialect
--- a/lib/ruby_smb/client/encryption.rb
+++ b/lib/ruby_smb/client/encryption.rb
@@ -4,18 +4,22 @@ module RubySMB
     module Encryption
       def smb3_encrypt(data)
         unless @client_encryption_key
+          key_bit_len = OpenSSL::Cipher.new(@encryption_algorithm).key_len * 8
+
           case @dialect
           when '0x0300', '0x0302'
             @client_encryption_key = RubySMB::Crypto::KDF.counter_mode(
               @session_key,
               "SMB2AESCCM\x00",
-              "ServerIn \x00"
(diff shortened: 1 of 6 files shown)


cdelafuente-r7:
Thanks @zeroSteiner ! It looks good to me. I tested against Windows 11 21H2 and verified it negotiated AES-256-GCM, AES-256-CCM, AES-128-GCM and AES-128-CCM. I also verified it still work with Windows 10 (both AES-128-GCM and AES-128-CCM).

However, the SMB Server seems to break when trying to parse a Negotiate request from Windows 11. It looks like two new [Negotiate Context](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/15332256-522e-4a53-8cd7-0bd17678a2f7) types have been added: [SMB2_RDMA_TRANSFORM_CAPABILITIES](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/52b74a74-9838-4f51-b2b0-efeb23bd79d6) and [SMB2_SIGNING_CAPABILITIES](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/cb9b5d66-b6be-4d18-aa66-8784a871cc10). Since this is out of scope, I'll create a specific ticket to fix this.

That said, I'll go ahead and land it.

cdelafuente-r7 (approved):

PR #212, opened 2022-04-12
Fix negotiation for SMB v2.0.2

zeroSteiner:
Some older Windows servers will respond back to an SMB 1 negotiate request with a dialect of 2.0.2 instead of the wildcard dialect. When that occurs, the client doesn't increment the message ID correctly which causes the subsequent session negotiation request to be rejected.

This issue can be reproduced with older versions of Server 2008. Once the issue is patched, the session setup should proceed correctly.

I opted to update the logic from doing an explicit check for the wildcard dialect (which was failing) to checking if an SMB2/3 response was received to an SMB1 request. This is a bit more flexible. I then tested it with servers running 2.0.2 (fixed), 2.1 and 3.1.1.

### DIFF
--- a/spec/lib/ruby_smb/client_spec.rb
+++ b/spec/lib/ruby_smb/client_spec.rb
@@ -1279,13 +1279,13 @@ RSpec.describe RubySMB::Client do
       end
 
       it 'calls the backing methods' do
-        expect(client).to receive(:negotiate_request).twice.and_call_original
+        request_packet = double('Request packet')
+        allow(client).to receive(:negotiate_request).and_return(request_packet)
+        allow(request_packet).to receive(:packet_smb_version)
+        expect(client).to receive(:negotiate_request)
         expect(client).to receive(:send_recv)
         expect(client).to receive(:negotiate_response)
-        expect(client).to receive(:parse_negotiate_response).twice do
-          client.dialect = '0x02ff'
-          client.smb1 = false
-        end
+        expect(client).to receive(:parse_negotiate_response)
         client.negotiate
       end
 
--- a/spec/lib/ruby_smb/client_spec.rb
+++ b/spec/lib/ruby_smb/client_spec.rb
@@ -1279,10 +1279,13 @@ RSpec.describe RubySMB::Client do
       end
 
       it 'calls the backing methods' do
-        expect(client).to receive(:negotiate_request)
+        expect(client).to receive(:negotiate_request).twice.and_call_original
         expect(client).to receive(:send_recv)
         expect(client).to receive(:negotiate_response)
-        expect(client).to receive(:parse_negotiate_response)
+        expect(client).to receive(:parse_negotiate_response).twice do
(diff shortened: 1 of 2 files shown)


adfoster-r7 (approved):
Fix works for me 🎉 

Before:
```
msf6 auxiliary(scanner/smb/smb_login) > run smb://Administrator:p4$$w0rd@192.168.123.133 lhost=192.168.123.1

[*] 192.168.123.133:445   - 192.168.123.133:445 - Starting SMB login bruteforce
[*] smb://Administrator:p4$$w0rd@192.168.123.133:445 - Error: 192.168.123.133: RubySMB::Error::CommunicationError An error occurred reading from the Socket Connection reset by peer
[*] smb://Administrator:p4$$w0rd@192.168.123.133:445 - Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

After:
```
msf6 auxiliary(scanner/smb/smb_login) > run smb://Administrator:p4$$w0rd@192.168.123.133

[*] 192.168.123.133:445   - 192.168.123.133:445 - Starting SMB login bruteforce
[+] 192.168.123.133:445   - 192.168.123.133:445 - Success: '.\Administrator:p4$$w0rd'
[*] smb://Administrator:p4$$w0rd@192.168.123.133:445 - Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
```

spec/lib/ruby_smb/client_spec.rb
```diff
@@ -1279,10 +1279,13 @@
...
+        expect(client).to receive(:negotiate_request).twice.and_call_original
         expect(client).to receive(:send_recv)
         expect(client).to receive(:negotiate_response)
-        expect(client).to receive(:parse_negotiate_response)
+        expect(client).to receive(:parse_negotiate_response).twice do
+          client.dialect = '0x02ff'
+          client.smb1 = false
+        end
         client.negotiate
       end
```
cdelafuente-r7, line 1290: It looks like this example now checks something else it was originally created for. I believe it was to make sure the backing methods were called (even if it is not that useful, I agree). Now, it is checking the logic when a wildcard is received, which is good, but I think this is what is done in the `with a wildcard revision number response` context below.

That would be great to have a specific set of examples that make sure it handles correctly response from old Windows (2.0.2 instead of wildcard) and it does not break the logic for other dialects 2.1 and 3.1.1, as you described. Note that I think the `with 0x0311 dialect` context below already validate the latter.

That said, this is how I would fix this example with the changes in `#negotiate`:
```ruby
      it 'calls the backing methods' do
        request_packet = double('Request packet')
        allow(client).to receive(:negotiate_request).and_return(request_packet)
        allow(request_packet).to receive(:packet_smb_version)
        expect(client).to receive(:negotiate_request)
        expect(client).to receive(:send_recv)
        expect(client).to receive(:negotiate_response)
        expect(client).to receive(:parse_negotiate_response)
        client.negotiate
      end
```

PR #214, opened 2022-04-18
Support guest authentication

zeroSteiner:
Support guest authentication. Per the specs a guest user is implementation-specific. This effectively allows incoming connections to authenticate when the user doesn't exist in the database. It's necessary to disable encryption in this case for SMB3 unfortunately.

> Otherwise, if the returned src_name corresponds to an implementation-specific guest user,[<270>](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/a64e55aa-1152-48e4-8206-edd96444e7f7#Appendix_A_270) the server MUST set the SMB2_SESSION_FLAG_IS_GUEST in the SessionFlags field of the SMB2 SESSION_SETUP Response and MUST set Session.IsGuest to TRUE.

See: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/5ed93f06-a1d2-4837-8954-fa8b833c2654

This setting is disabled by default. In the future it'll be necessary to support Metasploit hosting files for incoming requests which may be coming from an authenticated context.

# Testing

The example file server now has a `--guests` option. When enabled, incoming users will be authenticated using the guest context. Turn that option on, and do not add an explicit account.

- [ ] Run: `ruby examples/file_server.rb --path /var/public --share public --guest`
- [ ] Request a file from a Windows system: `type \\192.168.159.128\public\test.txt`
- [ ] See in the logging output that the NTLM authentication succeeded for a user and that it was marked as a guest as noted with the ` (guest)` suffix

### DIFF
--- a/lib/ruby_smb/gss/provider/ntlm.rb
+++ b/lib/ruby_smb/gss/provider/ntlm.rb
@@ -109,6 +109,7 @@ module RubySMB
           def process_ntlm_type3(type3_msg)
             if type3_msg.user == '' && type3_msg.domain == ''
               if @provider.allow_anonymous
+                @session_key = "\x00".b * 16 # see MS-NLMP section 3.4
                 return WindowsError::NTStatus::STATUS_SUCCESS
               end
 
@@ -122,6 +123,12 @@ module RubySMB
               domain: type3_msg.domain
             )
             if account.nil?
+              if @provider.allow_guests
+                logger.info("NTLM authentication request succeeded for #{dbg_string} (guest)")
+                @session_key = "\x00".b * 16 # see MS-NLMP section 3.4
+                return WindowsError::NTStatus::STATUS_SUCCESS
+              end
+
               logger.info("NTLM authentication request failed for #{dbg_string} (no account)")
               return WindowsError::NTStatus::STATUS_LOGON_FAILURE
             end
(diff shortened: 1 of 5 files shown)


lib/ruby_smb/gss/provider/ntlm.rb
```diff
@@ -229,25 +236,31 @@ def process_gss_type3(gss_api)
                 domain: type3_msg.domain
               )
               if account.nil?
-                if @provider.allow_anonymous
+                if type3_msg.user == ''
+                  is_guest = false
                   identity = IDENTITY_ANONYMOUS
+                else
+                  is_guest = true
+                  identity = Account.new(type3_msg.user.encode(''.encoding), '', type3_msg.domain.encode(''.encoding)).to_s
```
zeroSteiner, line 244: If the user wasn't specified, the identity is anonymous. If the user was specified and the account doesn't exist, it's a guest so use `Account#to_s` to format it properly.

cdelafuente-r7:
Thanks @zeroSteiner ! It looks good to me. I tested from a Windows 10 client with SMBv1, v2 and v3 and confirmed the user is correctly authenticated as the guest user. I'll go ahead and land it.

lib/ruby_smb/server/server_client/session_setup.rb
```diff
@@ -80,12 +80,14 @@ def do_session_setup_smb2(request, session)
           if gss_result.nt_status == WindowsError::NTStatus::STATUS_SUCCESS
             session.state = :valid
             session.user_id = gss_result.identity
+            session.is_guest = !!gss_result.is_guest
             session.key = @gss_authenticator.session_key
-            session.signing_required = request.security_mode.signing_required == 1
+            session.signing_required = request.security_mode.signing_required == 1 || (!session.is_guest && !session.is_anonymous)
```
gwillcox-r7, line 85: Will this not force signing if its not explicitly enabled but the session is not a guest or anonymous session even if the server doesn't support signing?

zeroSteiner: This is the server and the server supports signing. Signing is enabled if it's required or the user is not a guest or anonymous.

PR #215, opened 2022-04-19
Expose Socket Information

zeroSteiner:
This updates the ServerClient class to expose some additional information like the peer host and port. This can be used by something like Metasploit to more easily identify the client connection.

### DIFF
--- a/lib/ruby_smb/server/server_client.rb
+++ b/lib/ruby_smb/server/server_client.rb
@@ -19,7 +19,7 @@ module RubySMB
       include RubySMB::Server::ServerClient::ShareIO
       include RubySMB::Server::ServerClient::TreeConnect
 
-      attr_reader :dialect, :session_table
+      attr_reader :dialect, :dispatcher, :session_table
 
       # @param [Server] server the server that accepted this connection
       # @param [Dispatcher::Socket] dispatcher the connection's socket dispatcher
@@ -55,6 +55,14 @@ module RubySMB
       # @return [String]
       def getpeername
         @dispatcher.tcp_socket.getpeername
+      end
+
+      def peerhost
+        ::Socket::unpack_sockaddr_in(getpeername)[1]
+      end
+
+      def peerport
+        ::Socket::unpack_sockaddr_in(getpeername)[0]
       end
 
       #
--- a/spec/lib/ruby_smb/server/server_client_spec.rb
+++ b/spec/lib/ruby_smb/server/server_client_spec.rb
@@ -5,6 +5,7 @@ RSpec.describe RubySMB::Server::ServerClient do
   subject(:server_client) { described_class.new(server, dispatcher) }
 
   it { is_expected.to respond_to :dialect }
+  it { is_expected.to respond_to :dispatcher }
   it { is_expected.to respond_to :session_table }
 
   describe '#disconnect!' do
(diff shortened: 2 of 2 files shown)


lib/ruby_smb/server/server_client.rb
```diff
@@ -57,6 +57,14 @@ def getpeername
         @dispatcher.tcp_socket.getpeername
       end
 
+      def peerhost
+        ::Socket::unpack_sockaddr_in(getpeername)[1]
```
adfoster-r7, line 61: Looks like there's a the potential to get a NPE here:

```
 11 [04/21/2022 17:08:18] [e(0)] core: Exploit failed (windows/smb/group_policy_startup): NoMethodError undefined method `getpeername' for nil:NilClass - NoMethodError undefined method `getpeername' for nil:NilClass
 10 Call stack:
  9 /Users/user/.rvm/gems/ruby-3.0.2@metasploit-framework/bundler/gems/ruby_smb-203c90e2a354/lib/ruby_smb/server/server_client.rb:57:in `getpeername'
  8 /Users/user/.rvm/gems/ruby-3.0.2@metasploit-framework/bundler/gems/ruby_smb-203c90e2a354/lib/ruby_smb/server/server_client.rb:61:in `peerhost'
  7 /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/smb/server.rb:70:in `block in start_service'
  6 /Users/user/.rvm/gems/ruby-3.0.2@metasploit-framework/bundler/gems/ruby_smb-203c90e2a354/lib/ruby_smb/server.rb:66:in `block in run'
  5 /Users/user/.rvm/gems/ruby-3.0.2@metasploit-framework/bundler/gems/ruby_smb-203c90e2a354/lib/ruby_smb/server.rb:63:in `loop'
  4 /Users/user/.rvm/gems/ruby-3.0.2@metasploit-framework/bundler/gems/ruby_smb-203c90e2a354/lib/ruby_smb/server.rb:63:in `run'
  3 /Users/user/Documents/code/metasploit-framework/lib/msf/core/exploit/remote/smb/server.rb:20:in `block in start'
  2 /Users/user/Documents/code/metasploit-framework/lib/rex/thread_factory.rb:22:in `block in spawn'
  1 /Users/user/Documents/code/metasploit-framework/lib/msf/core/thread_manager.rb:105:in `block in spawn'
  0 /Users/user/.rvm/gems/ruby-3.0.2@metasploit-framework/gems/logging-2.3.0/lib/logging/diagnostic_context.rb:474:in `block in create_with_logging_context'
```

From the caller:

```ruby
        thread_factory = Proc.new do |server_client, &block|
          Rex::ThreadFactory.spawn("SMBServerClient(#{server_client.peerhost}->#{server_client.dispatcher.tcp_socket.localhost})", false, &block)
        end
```

I'm not sure what caused it just yet; but will keep an eye on exact replication steps 😄

gwillcox-r7 (approved):

gwillcox-r7:
Code looks good but will likely need to address the issue mentioned above.

lib/ruby_smb/server/server_client.rb
adfoster-r7: Unfortunately wasn't able to replicate again, but it looks like NPE is possible either way. Happy to leave this as-is, or to defensively allow this to handle the nil scenario safely :+1:

gwillcox-r7: Given this is likely an issue that will be encountered again in the future I'd be in favor of coding this defensively, as this sort of issue should be properly handled vs stack tracking. Also I can easily see this resulting in an issue tracker ticket for a confused end user.

I'd say it shouldn't be hard to handle this defensively and I'd make the change myself my only question would be what we should be returning in the case of an error here as I'm not 100% sure that returning an empty string here would be handled appropriately by calling code.

zeroSteiner: I'd be in favor of leaving it as is if we can't reproduce it. I'm pretty sure the stack trace is complaining that `@dispatcher.tcp_socket` is nil which I'm not sure how that's possible. If the `#tcp_socket` is nil, then nothing is going to work because the dispatcher is horribly broken.

Returning nil in this method on error seems reasonable enough but we'd just be kicking the problem down the road because the dispatcher is defunct and won't be able to send/recv.

gwillcox-r7: Sounds reasonable to me @zeroSteiner will get this landed now then 👍

PR #227, opened 2022-05-13
Move the server CLI into a library

zeroSteiner:
This adds the `RubySMB::Server::Cli` module with the code necessary to start a server from the command line interface. It consolidates all the duplicate options in the examples and will make it easy to add and smb server tool to metasploit in the future without a ton of duplicated code.

# Testing
- [ ] Run `examples/file_server.rb` with various options and configurations, see it start and honor the settings specified on the command line
- [ ] Do the same thing for `examples/virtual_file_server.rb`
- [ ] Read the docs, make sure they make sense

### DIFF
--- a/examples/file_server.rb
+++ b/examples/file_server.rb
@@ -10,7 +10,7 @@ Encoding.default_internal = 'UTF-8' if Encoding.default_internal.nil?
 options = RubySMB::Server::Cli.parse(defaults: { share_path: '.' }) do |options, parser|
   parser.banner = "Usage: #{File.basename(__FILE__)} [options]"
 
-  parser.on("--path PATH", "The path to share (default: #{options[:share_path]})") do |path|
+  parser.on("--share-path SHARE_PATH", "The path to share (default: #{options[:share_path]})") do |path|
     options[:share_path] = path
   end
 end
--- a/examples/file_server.rb
+++ b/examples/file_server.rb
@@ -7,7 +7,7 @@ require 'ruby_smb'
 # we just need *a* default encoding to handle the strings from the NTLM messages
 Encoding.default_internal = 'UTF-8' if Encoding.default_internal.nil?
 
-options = RubySMB::Server::Cli.parse!(defaults: { share_path: '.' }) do |options, parser|
+options = RubySMB::Server::Cli.parse(defaults: { share_path: '.' }) do |options, parser|
   parser.banner = "Usage: #{File.basename(__FILE__)} [options]"
 
   parser.on("--path PATH", "The path to share (default: #{options[:share_path]})") do |path|
--- a/examples/file_server.rb
+++ b/examples/file_server.rb
@@ -3,81 +3,19 @@
 require 'bundler/setup'
 require 'optparse'
 require 'ruby_smb'
-require 'ruby_smb/gss/provider/ntlm'
 
 # we just need *a* default encoding to handle the strings from the NTLM messages
 Encoding.default_internal = 'UTF-8' if Encoding.default_internal.nil?
 
-options = {
-  allow_anonymous: true,
-  allow_guests: false,
-  domain: nil,
-  username: 'RubySMB',
-  password: 'password',
-  share_name: 'home',
-  share_path: '.',
-  smbv1: true,
-  smbv2: true,
-  smbv3: true
-}
-OptionParser.new do |opts|
-  opts.banner = "Usage: #{File.basename(__FILE__)} [options]"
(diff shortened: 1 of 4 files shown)


lib/ruby_smb/server/cli.rb
```diff
@@ -0,0 +1,121 @@
...
+      #   * :allow_anonymous [Boolean] Whether or not to allow anonymous authentication.
+      #   * :allow_guest [Boolean] Whether or not to allow guest authentication.
+      #   * :username [String] The username of the account to add for authentication.
+      #   * :password [String] The password of the account to add for authentication.
+      #   * :domain [String] The domain of the account to add for authentication.
+      #   * :smbv1 [Boolean] Whether or not to enable SMBv1 dialects.
+      #   * :smbv2 [Boolean] Whether or not to enable SMBv2 dialects.
+      #   * :smbv3 [Boolean] Whether or not to enable SMBv3 dialects.
+      #   * :share_name [String] The name of the share to add.
+      def self.parse!(defaults: {}, &block)
```
dwelch-r7, line 34:
```suggestion
      def self.parse(defaults: {}, &block)
```

dwelch-r7 (approved):

lib/ruby_smb/server/cli.rb
```diff
@@ -0,0 +1,121 @@
...
+        )
+        ntlm_provider.put_account(options[:username], options[:password], domain: options[:domain])  # password can also be an NTLM hash
+
+        server = RubySMB::Server.new(
+          gss_provider: ntlm_provider,
+          logger: :stdout
+        )
+        server.dialects.select! { |dialect| RubySMB::Dialect[dialect].family != RubySMB::Dialect::FAMILY_SMB1 } unless options[:smbv1]
+        server.dialects.select! { |dialect| RubySMB::Dialect[dialect].family != RubySMB::Dialect::FAMILY_SMB2 } unless options[:smbv2]
+        server.dialects.select! { |dialect| RubySMB::Dialect[dialect].family != RubySMB::Dialect::FAMILY_SMB3 } unless options[:smbv3]
```
adfoster-r7, line 97: Not a blocker, this might read better as `filter!`

```suggestion
        server.dialects.filter! { |dialect| RubySMB::Dialect[dialect].family != RubySMB::Dialect::FAMILY_SMB3 } unless options[:smbv3]
```

lib/ruby_smb/server/cli.rb
```diff
@@ -0,0 +1,121 @@
...
+
+        server
+      end
+
+      # Run the server forever. At least 1 SMB dialect must be enabled.
+      #
+      # @param [RubySMB::Server] server The server instance to run.
+      # @param [#puts] out The stream to write standard output messages to.
+      # @param [#puts] err The stream to write error messages to.
+      def self.run(server, out: $stdout, err: $stderr)
```
adfoster-r7, line 107: Not a blocker, I think the logger abstraction that was introduced could potentially be used here

lib/ruby_smb/server/cli.rb
```diff
@@ -0,0 +1,121 @@
...
+              options[:username] = username
+            end
+          end
+
+          parser.on("--password PASSWORD", "The account's password (default: #{defaults[:password]})") do |password|
+            options[:password] = password
+          end
+
+          block.call(options, parser) if block_given?
+        end.parse!
```
adfoster-r7, line 75: Does this need to be `parse` as well? 👀

examples/file_server.rb
```diff
@@ -3,81 +3,19 @@
...
-  smbv2: true,
-  smbv3: true
-}
-OptionParser.new do |opts|
-  opts.banner = "Usage: #{File.basename(__FILE__)} [options]"
-  opts.on("--path PATH", "The path to share (default: #{options[:share_path]})") do |path|
+options = RubySMB::Server::Cli.parse(defaults: { share_path: '.' }) do |options, parser|
+  parser.banner = "Usage: #{File.basename(__FILE__)} [options]"
+
+  parser.on("--path PATH", "The path to share (default: #{options[:share_path]})") do |path|
```
adfoster-r7, line 13: What are your thoughts on prefixing these for more clarity? 🤔 

```
--share-name
--share-path
```

lib/ruby_smb/server/cli.rb
zeroSteiner: No, if I change that from `#parse!` to `#parse`, the option parser stops working in the same way.

PR #233, opened 2022-06-30
Add a couple of MS-DFSNM definitions and functions

zeroSteiner:
This adds the MS-DFSNM definitions and functions to implement the Dfscoerce attack which is similar to petitpotam.

# Testing
- [x] Make sure the unit tests pass
- [x] Use the Metasploit module with both `METHOD` options (one for each function implemented here)

Looking through the documentation for [NetrDfsAddStdRoot](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-dfsnm/b18ef17a-7a9c-4e22-b1bf-6a4d07e87b2d), I was unable to get the function to work with a nil / null pointer comment. From what I can tell the comment *must* be specified as a string. Also the parameter types are all marked as `WCHAR*` but the invocation fails when the types are defined as `ndr_wide_stringz_ptr`.

### DIFF
--- a/lib/ruby_smb/dcerpc.rb
+++ b/lib/ruby_smb/dcerpc.rb
@@ -46,6 +46,7 @@ module RubySMB
     require 'ruby_smb/dcerpc/epm'
     require 'ruby_smb/dcerpc/drsr'
     require 'ruby_smb/dcerpc/sec_trailer'
+    require 'ruby_smb/dcerpc/dfsnm'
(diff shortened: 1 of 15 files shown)


space-r7 (commented):
Looks good to me. All specs passed

space-r7:
Specs passed and tested successfully with the module. Will go ahead and get this landed.

PR #236, opened 2022-08-23
Add Initial MS-ICPR definitions

zeroSteiner:
This adds the initial implementation for MS-ICPR which deals with certificates from Active Directory. There's only one function implemented at this time and that's `cert_server_request` which allows the caller to request, retrieve and renew certificates. It's up to the caller to properly generate the Certificate Signing Request (CSR).

That's all relatively basic stuff that follows existing DCERPC API patterns. The bulk of the changes in this PR are actually what's necessary to make DCERPC calls over named pipes using `RPC_C_AUTHN_LEVEL_PKT_PRIVACY`. Until now, this feature was only available to the raw DCERPC client. While making `cert_server_request` calls, it became apparent that the feature was also needed when invoked over a named pipe, regardless of SMB 3 encryption.

The code was largely moved from the `Dcerpc::Client` class and into the `Dcerpc` module which `Dcerpc::Client` already included. After that, it was necessary to tweak the `SMB#::Pipe` classes to properly encapsulate and handle dcerpc requests when `RPC_C_AUTHN_LEVEL_PKT_PRIVACY` is in use.

The `auxiliary/gather/windows_secrets_dump` module is the best way to test this since it uses both named pipes and the DCERPC client. Running that module should show that the changes don't break compatibility at all. To show that the new functionality is working, it'll be required by a module that will be PRed to Metasploit pretty soon.

## Testing

- [ ] Run `auxiliary/gather/windows_secrets_dump` against a DC, see that it works correctly without any errors demonstrating that everything is still working
- [ ] See the unit tests all pass
- [ ] Test the new functionality with the to-be-released module (will update this then)


lib/ruby_smb/dcerpc.rb
```diff
@@ -55,35 +56,267 @@ module Dcerpc
...
+    end
+
+    # Send a packet to the remote host
+    #
+    # @param packet [BinData::Record] the packet to send
+    # @raise [Error::CommunicationError] if socket-related error occurs
+    def send_packet(packet)
+      write(data: packet.to_binary_s)
+      nil
+    end
```
zeroSteiner, line 157: These three functions were added for API compatibility so the other methods could be consolidated here.

lib/ruby_smb/dcerpc/ndr.rb
```diff
@@ -1208,14 +1208,50 @@ class NdrWideStringzPtr < NdrConfVarWideStringz
...
+    new_klass_name = "Ndr#{klass}ConfArrayPtr"
+    unless self.const_defined?(new_klass_name)
+      new_klass = Class.new(NdrConfArray) do
+        default_parameters type: "ndr_#{klass}".downcase.to_sym
+        extend PointerClassPlugin
+      end
+      self.const_set(new_klass_name, new_klass)
+      BinData::RegisteredClasses.register(new_klass_name, new_klass)
+    end
   end
```
zeroSteiner, line 1255: This defines additional arrays. While working on the function's definitions, I had to find the correct array definition type and used this. The Byte one is the same, but allows assigning a string and will call `#bytes` on it for convenience.

lib/ruby_smb/smb1/pipe.rb
```diff
@@ -30,10 +30,18 @@ def initialize(tree:, response:, name:)
...
           extend RubySMB::Dcerpc::Dfsnm
+        when 'cert', '\\cert'
+          extend RubySMB::Dcerpc::Icpr
         end
         super(tree: tree, response: response, name: name)
       end
 
+      def bind(options={})
+        @size = 1024
+        @ntlm_client = @tree.client.ntlm_client
```
zeroSteiner, line 41: These are pipe-specific instance variables that need to be defined.

cdelafuente-r7 (commented):
Thanks @zeroSteiner for this PR! I just left a couple of minor comments, but otherwise, it looks good to me. I tested with the related MSF [module](https://github.com/rapid7/metasploit-framework/pull/16939) and make sure the `gather/windows_secrets_dump` module still works as expected.

cdelafuente-r7:
Thanks @zeroSteiner ! Everything looks good now. I'll go ahead and land it.

PR #239, opened 2022-11-02
Fix DCERPC Encrypted Fragments

zeroSteiner:
This makes two changes.

The first fixes an error where when encryption is used, only the first fragment of a fragmented response would be decrypted. This means the rest of the data would be corrupt which leads to weird parsing errors. Those weird parsing errors are what I was debugging and I originally thought the issue was in the BinData definitions, so while digging into that I added tracing to all of the direct IO read operations in `dcerpc/ndr.rb` which is the second change. Now if the user enables bindata read tracing to do what I was doing, they'll get additional information like what the ref ID, and max_count files are being set to.

The tracing output was pretty helpful because it made it obvious that the size field was incorrect, it just wasn't BinData's fault in my case.

## Example Tracing Output

```
BinData.trace_reading { RubySMB::Dcerpc::Icpr::CertServerRequestResponse.read(File.binread('/tmp/response_good.bin')) }
obj.pdw_request_id => 59
obj.pdw_disposition => 3
obj.pctb_cert.cb => 2662
obj.pctb_cert.pb.ref_id => 131072
obj.pctb_cert.pb.max_count => 2662
obj.pctb_cert.pb[0] => 48
obj.pctb_cert.pb[1] => 130
obj.pctb_cert.pb[2] => 10
obj.pctb_cert.pb[3] => 98
```

### DIFF
--- a/lib/ruby_smb/dcerpc/ndr.rb
+++ b/lib/ruby_smb/dcerpc/ndr.rb
@@ -247,12 +247,22 @@ module RubySMB::Dcerpc::Ndr
 
     def do_read(io)
       if is_a?(ConfPlugin) && should_process_max_count?
-        set_max_count(io.readbytes(4).unpack('L<').first)
+        max_count = io.readbytes(4).unpack1('L<')
+        BinData.trace_message do |tracer|
+          tracer.trace_obj("#{debug_name}.max_count", max_count.to_s)
+        end
+        set_max_count(max_count)
       end
 
       if is_a?(VarPlugin)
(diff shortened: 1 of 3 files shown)


cdelafuente-r7 (approved):

cdelafuente-r7:
Thanks @zeroSteiner for fixing this. I could not replicate the issue using the MSF `icpr_cert` module but these changes look good to me. I tested with both `icpr_cert` and `windows_secrets_dump` modules and verify they're still working normally. I'll go ahead and land it.

PR #242, opened 2022-12-02
Add NDR Type Serialization Version 1 structures

cdelafuente-r7:
This adds the NDR `Type Serialization Version 1` structures as defined [here](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/9a1d0f97-eac0-49ab-a197-f1a581c2d6a0).

These structures are referred as `Pickles` by the NDR protocol (see [(IDL/NDR) # Pickles](https://pubs.opengroup.org/onlinepubs/9668899/chap2.htm#tagcjh_05_01_07)). These structures are needed by this Metasploit [module](https://github.com/rapid7/metasploit-framework/pull/17066) and already implemented [here](https://github.com/cdelafuente-r7/metasploit-framework/blob/1e5b13581711746b089c4b41a1e2d9e70524ef31/lib/rex/proto/kerberos/pac/credential_info.rb#L3-L34). They should be moved to RubySMB, since they are DCERPC NDR structures.

### DIFF
--- a/lib/ruby_smb/dcerpc/ndr.rb
+++ b/lib/ruby_smb/dcerpc/ndr.rb
@@ -1315,5 +1315,40 @@ module RubySMB::Dcerpc::Ndr
     end
   end
 
+  # (IDL/NDR) Pickles as defined in
+  # [(IDL/NDR) # Pickles](https://pubs.opengroup.org/onlinepubs/9668899/chap2.htm#tagcjh_05_01_07)
+  # and
+  # [2.2.6 Type Serialization Version # 1](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/9a1d0f97-eac0-49ab-a197-f1a581c2d6a0)
+
+  # [2.2.6.1 Common Type Header for the Serialization Stream](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/6d75d40e-e2d2-4420-b9e9-8508a726a9ae)
+  class TypeSerialization1CommonTypeHeader < NdrStruct
+    default_parameter byte_align: 8
+    endian :little
+
+    ndr_uint8  :version, initial_value: 1
+    ndr_uint8  :endianness, initial_value: 0x10
+    ndr_uint16 :common_header_length, initial_value: 8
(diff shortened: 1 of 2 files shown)


lib/ruby_smb/dcerpc/ndr.rb
```diff
@@ -1315,5 +1315,40 @@ def set(handle)
...
+    ndr_uint16 :common_header_length, initial_value: 8
+    ndr_uint32 :filler, initial_value: 0xCCCCCCCC
+  end
+
+  # [2.2.6.2 Private Header for Constructed Type](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/63949ba8-bc88-4c0c-9377-23f14b197827)
+  class TypeSerialization1PrivateHeader < NdrStruct
+    default_parameter byte_align: 8
+    endian :little
+
+    ndr_uint32 :object_buffer_length
```
adfoster-r7, line 1339: cc @dwelch-r7 I think you modified this a bit for your PAC model definitions - is there anything here that you think needs changed? 👀 

https://github.com/rapid7/metasploit-framework/blob/fe1f2bec1a9759f95e423a9ec48fbacedca0a266/lib/rex/proto/kerberos/pac/krb5_pac.rb#L6-L40

dwelch-r7: Yes I do, thanks for calling it out, I need a way to pass in the buffer length when creating a PAC structure, I used mandatory_parameters but if there's a more bindata-y way of doing it that's fine by me

adfoster-r7 (approved):
Latest changes looks good to me :+1: 

Not a blocker for me, since this code temporarily lives in [metasploit](https://github.com/rapid7/metasploit-framework/blob/a80db73bab0ee43f6ec3f1e261423b5c9ecf246f/lib/rex/proto/kerberos/ndr/type_serialization1.rb) we could probably get this merged in after the Kerberos feature branch is released to avoid extra overhead on maintaining the version bumps

smcintyre-r7:
The changes look the same as the code that is in Metasploit. To really test this we'll need to remove those definitions to ensure these are the ones used. I'll get this merged then start that process to double check.

PR #244, opened 2023-01-18
Don't crash when the buffer is empty

zeroSteiner:
When the response status is 'submitted', the certificate buffer may be empty. In that case, don't load the certificate because it will crash.

There will need to be a Metasploit PR to handle cases where the result is `:submitted` and the certificate is absent.

### DIFF
--- a/lib/ruby_smb/dcerpc/icpr.rb
+++ b/lib/ruby_smb/dcerpc/icpr.rb
@@ -51,6 +51,7 @@ module RubySMB
         end
 
         ret = {
+          certificate: nil,
           disposition: cert_server_request_response.pdw_disposition.value,
           disposition_message: cert_server_request_response.pctb_disposition_message.buffer.chomp("\x00\x00").force_encoding('utf-16le').encode,
           status: {
@@ -72,7 +73,7 @@ module RubySMB
               status_code: status_code
             )
           end
-        else
+        elsif !cert_server_request_response.pctb_encoded_cert.buffer.empty?
           ret[:certificate] = OpenSSL::X509::Certificate.new(cert_server_request_response.pctb_encoded_cert.buffer)
         end
 


lib/ruby_smb/dcerpc/icpr.rb
```diff
@@ -72,7 +72,7 @@ def cert_server_request(attributes:, authority:, csr:)
               status_code: status_code
             )
           end
-        else
+        elsif !cert_server_request_response.pctb_encoded_cert.buffer.empty?
```
gwillcox-r7, line 75: Doesn't look like we are checking if `cert_server_request_response` is `nil` here? Not sure if that check should be moved to be earlier in the code though. Same also applies to `pctb_encoded_cert` as well.

Other than that this looks good!

zeroSteiner: It won't be nil, it's defined on L48 and if the read fails it's raised as an InvalidPacket error already. If `cert_server_request_response` is defined, then the `pctb_encoded_cert` attribute is defined because that's in the BinData definition.

lib/ruby_smb/dcerpc/icpr.rb
```diff
@@ -72,7 +72,7 @@ def cert_server_request(attributes:, authority:, csr:)
               status_code: status_code
             )
           end
-        else
```
bwatters-r7, line 75: Super minor, but based on this and the definition for `ret` on L, `ret` will not contain the key `certificate`.  I'm assuming that's expected, though typically it makes me feel warmer and fuzzier to have `ret[:certificate]` be `nil` rather than undefined.  I don't know if we want to add the `certificate` key to the original `ret` definition, or if we always check or catch the key exception when we try to access it.

zeroSteiner: Yeah the `:certificate` key won't be set when the status is error, so callers should already account for certificate only being present sometimes. I can get it updated so `:certificate` is initialized in `ret` to be nil though. 👍🏻

PR #245, opened 2023-01-18
Update ruby versions

adfoster-r7:


smcintyre-r7:
This is an actions specific problem. I see that the tests are running correctly now so I've gone ahead and merged this fix in. Thanks alot @adfoster-r7 !

PR #248, opened 2023-02-22
Add definitions for three new EFS RPC requests and responses

zeroSteiner:
This adds definitions for the following three EFS RPC methods. They'll be required by updates to the petitpotam module I will submit to Metasploit soon.

* [EfsRpcDecryptFileSrv](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/043715de-caee-402a-a61b-921743337e78)
* [EfsRpcQueryRecoveryAgents](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/cf759c00-1b90-4c33-9ace-f51c20149cea)
* [EfsQueryUsersOnFile](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/a058dc6c-bb7e-491c-9143-a5cb1f7e7cea)

The easiest way to test this will be with the new petitpotam module changes which will add each of these in as a new method that can be selected. More detailed steps will be posted in that PR.

### DIFF
--- a/lib/ruby_smb/dcerpc/encrypting_file_system.rb
+++ b/lib/ruby_smb/dcerpc/encrypting_file_system.rb
@@ -82,6 +82,8 @@ module RubySMB
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_open_file_raw_response'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_query_recover_agents_request'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_query_recover_agents_response'
+      require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_query_users_on_file_request'
+      require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_query_users_on_file_response'
     end
   end
 end
--- a/lib/ruby_smb/dcerpc/encrypting_file_system.rb
+++ b/lib/ruby_smb/dcerpc/encrypting_file_system.rb
@@ -35,12 +35,53 @@ module RubySMB
       OVERWRITE_HIDDEN = 0x00000004
       EFS_DROP_ALTERNATE_STREAMS = 0x00000010
 
+      class EfsHashBlob < Ndr::NdrStruct
+        endian :little
+        default_parameter byte_align: 4
+
+        ndr_uint32   :cb_data
+        ndr_byte_conf_array_ptr :b_data
+      end
+
+      class EfsHashBlobPtr < EfsHashBlob
+        extend Ndr::PointerClassPlugin
+      end
+
+      class EncryptionCertificateHash < Ndr::NdrStruct
+        endian :little
+        default_parameter byte_align: 4
+
+        ndr_uint32                :cb_total_length
+        prpc_sid                  :user_sid
+        efs_hash_blob_ptr         :certificate_hash
+        ndr_wide_stringz_ptr      :lp_display_information
+      end
+
+      class EncryptionCertificateHashPtr < EncryptionCertificateHash
+        default_parameter byte_align: 4
+        extend Ndr::PointerClassPlugin
+      end
+
+      class EncryptionCertificateHashList < BinData::Record
+        endian :little
+        default_parameter byte_align: 4
+
+        uint32 :ncert_hash
+        ndr_var_array :users, type: :encryption_certificate_hash_ptr
+      end
+
+      class EncryptionCertificateHashListPtr < EncryptionCertificateHashList
+        extend Ndr::PointerClassPlugin
+      end
+
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_decrypt_file_srv_request'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_decrypt_file_srv_response'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_encrypt_file_srv_request'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_encrypt_file_srv_response'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_open_file_raw_request'
       require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_open_file_raw_response'
+      require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_query_recover_agents_request'
+      require 'ruby_smb/dcerpc/encrypting_file_system/efs_rpc_query_recover_agents_response'
     end
   end
 end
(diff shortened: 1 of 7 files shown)


lib/ruby_smb/dcerpc/encrypting_file_system.rb
```diff
@@ -35,10 +35,57 @@ module EncryptingFileSystem
...
+        extend Ndr::PointerClassPlugin
+      end
+
+      # [2.2.11 ENCRYPTION_CERTIFICATE_HASH_LIST](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/2718804c-6ab9-45fd-98cf-541bc3b6bc75)
+      class EncryptionCertificateHashList < BinData::Record
+        endian :little
+        default_parameter byte_align: 4
+
+        uint32 :ncert_hash
+        ndr_var_array :users, type: :encryption_certificate_hash_ptr
```
cdelafuente-r7, line 72: According to the [specs](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-efsr/2718804c-6ab9-45fd-98cf-541bc3b6bc75), this fields seems to be a pointer:
```
A pointer to an array of pointers to ENCRYPTION_CERTIFICATE_HASH (2.2.10) structures.
```

I think this field is a pointer to a conformant array instead of a varying array. This is usually the case when the MIDL definition uses the `size_is()` [attribute](https://learn.microsoft.com/en-us/windows/win32/midl/size-is):
```
   [size_is(nCert_Hash,)] ENCRYPTION_CERTIFICATE_HASH** Users;
```

This attribute controls the amount of memory allocated for the data, which correspond to the NDR [conformant array](https://pubs.opengroup.org/onlinepubs/9629399/chap14.htm#tagcjh_19_03_03_02) `max_count` value.

I believe using a varying array here still works since he number of bytes is equal in both cases:
- pointer to conformant array: `[ uint32 ref_id ] [ uint32 max_count ] [ element 1 ] [ ... ]`
- varying array: `[ uint32 offset ] [ uint32 actual_count ] [ element 1 ] [ ... ]`

I haven't tested it yet, but one way to make sure my assumption is correct, is to inspect a packet that comes from a proper Windows host, which contains this structure. The second `uint32` will always be the size of the array wether it is conformant or a varying. However, the first `uint32` is usually 0x00000000 in case of a varying array (`offset`) and a correct `ref_id` for a pointer to a conformant array (usually 0x0002000, 0x0002004, 0x0002008, etc.).

Note that a pointer to a conformant array have specific sets of rules defined by the NDR syntax, which are different than a simple varying array when they are inserted in other constructed types like structures or arrays. It is not the case here, but if we want to reuse it, this can have some unwanted side effects.

cdelafuente-r7 (commented):
Thanks  for adding these structures @zeroSteiner ! I just left one comment about a possible mismatch between a pointer to a conformant array and a varying array field. Other than that, it looks good to me.

lib/ruby_smb/dcerpc/encrypting_file_system.rb
zeroSteiner: I see, yeah I think you're right. Nice catch I'll get this switched over.

For reference, I was using this script for test parsing an empty and populated response:

<details>
<summary>test_efs.rb</summary>

```
#!/usr/bin/ruby

# This script tests a full Authentication/Session Setup cycle
# including protocol negotiation and authentication.

require 'bundler/setup'
require 'ruby_smb'

empty = "\x00\x00\x00\x005\x00\x00\x00".b
# file_name: "\\\\localhost\\C$\\Users\\smcintyre\\Desktop\\Encrypted\\hello world.txt"
populated =  "\x00\x00\x02\x00".b
populated << "\x01\x00\x00\x00".b
populated << "\x04\x00\x02\x00\x01\x00\x00\x00\b\x00\x02\x00 \x00\x00\x00\x00\x00\x00\x00\f\x00\x02\x00\x14\x00\x02\x00\x14\x00\x00\x00\x10\x00\x02\x00\x14\x00\x00\x00\x15\xC9\xADB\x8A\x8Au\x10Lk\xDC}7v\n\xA1\xBD{\x00/\"\x00\x00\x00\x00\x00\x00\x00\"\x00\x00\x00s\x00m\x00c\x00i\x00n\x00t\x00y\x00r\x00e\x00(\x00s\x00m\x00c\x00i\x00n\x00t\x00y\x00r\x00e\x00@\x00m\x00s\x00f\x00l\x00a\x00b\x00.\x00l\x00o\x00c\x00a\x00l\x00)\x00\x00\x00\x00\x00\x00\x00".b

$stderr.puts 'empty:'
BinData::trace_reading do
    empty = RubySMB::Dcerpc::EncryptingFileSystem::EfsRpcQueryRecoveryAgentsResponse.read(empty)
end
$stderr.puts empty.inspect

$stderr.puts 'populated:'
BinData::trace_reading do
    populated = RubySMB::Dcerpc::EncryptingFileSystem::EfsRpcQueryRecoveryAgentsResponse.read(populated)
end
$stderr.puts populated.inspect
```

With your changes implemented, the output does make more sense:

```
...
populated:
obj.recover_agents.ref_id => 131072
obj.recover_agents.ncert_hash => 1
obj.recover_agents.users.ref_id => 131076
obj.recover_agents.users.max_count => 1
obj.recover_agents.users[0].ref_id => 131080
obj.recover_agents.users[0].cb_total_length => 32
obj.recover_agents.users[0].user_sid.ref_id => 0
obj.recover_agents.users[0].certificate_hash.ref_id => 131084
...
```

vs this with the offset field that was confusing me.
```
...
populated:
obj.recover_agents.ref_id => 131072
obj.recover_agents.ncert_hash => 1
obj.recover_agents.users.offset => 131076
obj.recover_agents.users.actual_count => 1
obj.recover_agents.users[0].ref_id => 131080
obj.recover_agents.users[0].cb_total_length => 32
obj.recover_agents.users[0].user_sid.ref_id => 0
obj.recover_agents.users[0].certificate_hash.ref_id => 131084
...
```
</details>

cdelafuente-r7: Absolutely, it makes sense. The pointer's `ref_id`'s now follow the expected pattern: 131072, 131076, 131080 and 131084 (in hex: 0x00020000, 0x00020004, 0x00020008, 0x0002000C). Thanks for looking into this.

cdelafuente-r7:
Thanks for updating this. I tested using the new PetitPotam module changes and everything looks good. I'll go ahead and land it.

cdelafuente-r7:
# Release Notes

This adds definitions for the `EfsRpcDecryptFileSrv`, EfsRpcQueryRecoveryAgents and `EfsQueryUsersOnFile` EFS RPC methods. They'll be required by the Metasploit PetitPotam module.

PR #250, opened 2023-03-22
virtual_disk: Add 'pathname' require

thesamesam:
Needed for join()'s call to Pathname.new(). Only comes up if disabling simplecov because of a transitive-require.


cdelafuente-r7:
Thanks for updating this. Would you mind sharing the steps to reproduce this issue? I used the example [script](https://github.com/rapid7/ruby_smb/blob/master/examples/virtual_file_server.rb) and was not able to reproduce the problem with simplecov disabled.

smcintyre-r7:
Closed since we can't reproduce the problem and have not received a response.

PR #256, opened 2023-11-08
Add support for callback hooks in share providers

zeroSteiner:
This adds the ability for developers to install callback hooks for share providers. The hooks are stored in the share provider instance but are actually invoked in the share processor. At this time, only one hook for a request class, and location combination is supported at a time but we could easily change that in the future. The location is "before" or "after" to allow the developer to hook the logic before or after the original handler is invoked. This can be used, for example, to monitor a connection for individual requests to track state.

There will be another pull request which will require these changes and can be used for testing.


bwatters-r7:
# Release Notes
This adds the ability for developers to install callback hooks for share providers.

PR #257, opened 2023-12-04
Close the opened directory after listing it's contents

dwelch-r7:
Noticed an issue when working on https://github.com/rapid7/metasploit-framework/pull/18539 where when listing the contents of a directory we don't close the handle to it resulting in weirdness on windows like not being able to rename any files located in that directory

To replicate add:
```
require 'pry-byebug'; binding.pry
puts 'done'
```
to the end of `examples/list_directory` to pause execution of the script so we don't disconnect
run this command `bundle exec ruby list_directory.rb <ip> <user> <pass> <share>`  example: `bundle exec ruby list_directory.rb 172.16.158.159 vagrant vagrant foo`

Then try to rename a file in the share you're listing and you'll get an error like this:

image

Check out this PR and run through the same steps and you should no longer hit the issue

### DIFF
--- a/lib/ruby_smb/smb2/tree.rb
+++ b/lib/ruby_smb/smb2/tree.rb
@@ -85,6 +85,7 @@ module RubySMB
       # @raise [RubySMB::Error::InvalidPacket] if the response is not a QueryDirectoryResponse packet
       def list(directory: nil, pattern: '*', type: RubySMB::Fscc::FileInformation::FileIdFullDirectoryInformation)
         create_response = open_directory(directory: directory)
+        opened_directory = RubySMB::SMB2::File.new(tree: self, response: create_response, name: directory)
         file_id         = create_response.file_id
 
         directory_request                         = RubySMB::SMB2::Packet::QueryDirectoryRequest.new
@@ -127,7 +128,7 @@ module RubySMB
           # Reset the message id so the client can update appropriately.
           directory_request.smb2_header.message_id = 0
         end
-
+        opened_directory.close
         files
       end
 
--- a/lib/ruby_smb/smb2/tree.rb
+++ b/lib/ruby_smb/smb2/tree.rb
@@ -128,8 +128,9 @@ module RubySMB
           # Reset the message id so the client can update appropriately.
           directory_request.smb2_header.message_id = 0
         end
-        opened_directory.close
         files
+      ensure
+        opened_directory.close
       end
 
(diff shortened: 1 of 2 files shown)


lib/ruby_smb/smb2/tree.rb
```diff
@@ -127,7 +128,7 @@ def list(directory: nil, pattern: '*', type: RubySMB::Fscc::FileInformation::Fil
           # Reset the message id so the client can update appropriately.
           directory_request.smb2_header.message_id = 0
         end
-
+        opened_directory.close
```
adfoster-r7, line 131: Does this need an ensure block, and error handling if the close fails? 🤔

adfoster-r7:
Does this mean there's potentially other resources being opened that aren't closed correctly? 🤔

lib/ruby_smb/smb2/tree.rb
dwelch-r7: putting it in an `ensure` makes sense to me done that, error handling if the close fails uuhhhh ideally yes? but I'm not sure what you would actually want to happen there, we could retry? I wouldn't want to bomb out if we can't close it because we'd still get what we wanted which in this case is the list of files in the directory 🤷

lib/ruby_smb/smb2/tree.rb
```diff
@@ -127,8 +128,9 @@ def list(directory: nil, pattern: '*', type: RubySMB::Fscc::FileInformation::Fil
           # Reset the message id so the client can update appropriately.
           directory_request.smb2_header.message_id = 0
         end
-
         files
+      ensure
+        opened_directory.close
```
adfoster-r7, line 133:
```suggestion
        opened_directory.close if opened_directory
```

adfoster-r7:
This seems legit to me; Will await an eyeball from @cdelafuente-r7 / @zeroSteiner incase there's a cleaner close API available here that should be used

cdelafuente-r7:
It looks good to me.

A another way would be to refactor this and move everything related to directory to a separate `RubySMB::SMB2::Directory` class or under the existing `RubySMB::SMB2::File` class. It looks like `open_directory` is not consistent with the other `open_file` and `open_pipe` methods. This one returns a response and the two others return a proper object (`RubySMB::SMB2::File` and `RubySMB::SMB2::Pipe`). The `list` method should also not be part of the `RubySMB::SMB2::Tree` class, IMHO.

That said, it is out-of-scope and this proposed fix is good to me.

PR #259, opened 2023-12-15
Update and refactor Efsrpc

zeroSteiner:
This makes a few changes that will be required by a forthcoming PR to the Metasploit Framework's existing petitpotam module.

* 5377044 -- This fixes a bug where frozen strings in the endpoint definitions would cause a crash. The update changes it so the `uuid_string` argument is not mutated.
* 45ad5be -- This makes some changes so that RubySMB's named pipe can be used for Efsrpc requests. The `Request` class needed to be updated so the `#dcerpc_request` method would accept the requests.
* 9a64a6d -- This adds the minimum definition to allow connecting to the LSARPC endpoint by defining the UUID and version constants.
* 9591eb6 -- This refactors the `EncryptingFileSystem` module to be renamed `Efsrpc` which makes it more consistent with the other DCERPC module that are named after the well-known pipe a user would connect to. See the definitions in [pipe.rb](https://github.com/rapid7/ruby_smb/blob/285716d3fad5ca9a14bb7cf9026fe6b756808bc4/lib/ruby_smb/smb2/pipe.rb#L16) for an example of how most modules are named after the pipe name.


adfoster-r7:
Thanks!

PR #260, opened 2024-01-08
Add the #remove_share method to the server

zeroSteiner:
This is an easy one. It just adds the `#remove_share` method to the server. I'll need to use it to clean up once a single server instance can be shared by multiple modules in Metasploit as an HTTP service can be.

For testing, the easiest thing would probably be to wait until the corresponding Metasploit PR has been made and then just check that the cleanup method used it to remove the share using `smbclient` (e.g. `smbclient //192.168.159.128/Share1 -U smcintyre -N -c "dir *"`).

### DIFF
--- a/lib/ruby_smb/server.rb
+++ b/lib/ruby_smb/server.rb
@@ -58,6 +58,14 @@ module RubySMB
       @shares[share_provider.name] = share_provider
     end
 
+    def remove_share(share_provider)
+      share_provider = share_provder.name if share_provider.is_a?(RubySMB::Server::Share::Provider::Base)
+      logger.debug("Removing share: #{share_provider}")
+      @shares.delete(share_provider)
+
+      nil
+    end
+
     # Run the server and accept any connections. For each connection, the block will be executed if specified. When the
     # block returns false, the loop will exit and the server will no long accept new connections.
     def run(&block)
--- a/lib/ruby_smb/server.rb
+++ b/lib/ruby_smb/server.rb
@@ -53,13 +53,15 @@ module RubySMB
       }
     end
 
+    # @param [Share::Provider::Base] share_provider the share provider to register
     def add_share(share_provider)
       logger.debug("Adding #{share_provider.type} share: #{share_provider.name}")
       @shares[share_provider.name] = share_provider
     end
 
+    # @param [String, Share::Provider::Base] share_provider the share provider to deregister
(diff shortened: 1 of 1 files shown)


lib/ruby_smb/server.rb
```diff
@@ -58,6 +58,14 @@ def add_share(share_provider)
       @shares[share_provider.name] = share_provider
     end
 
+    def remove_share(share_provider)
+      share_provider = share_provder.name if share_provider.is_a?(RubySMB::Server::Share::Provider::Base)
```
bwatters-r7, line 62: Two things.....
One, this looks like a typo?
```suggestion
      share_provider = share_proivder.name if share_provider.is_a?(RubySMB::Server::Share::Provider::Base)
```

Two, this looks like we're taking an element of an object and overwriting that object with the object's element?
On second thought, I see.  I'm guessing that `@shares.delete()` takes a string that is the `share_provider.name` value, but this line will overwrite the parameter to the `share_provider.name` value if the user passes in the `share_provider` object.
Could we maybe add a different variable name for this like `share_provider_name`, or even `share_provider_to_delete`?  Maybe even just a comment that this can take either the name or the `share_provider` object?

zeroSteiner: 1. Yes, that's a typo and I'll fix it.
2. I'm confused as to what you're saying. Basically `share_provider` needs to be the share name to remove. This line only serves to obtain that value from the `#name` attribute if the `share_provider` is an instance of `RubySMB::Server::Share::Provider::Base`, otherwise it needs to be a string. I'll add some docs to clarify that.

bwatters-r7: Per #2- It just struck me as odd we were renaming an object to an objects attribute.  I think the crux was that you're leveraging weak typing here as a feature to create a method that is designed to take multiple parameter types and deal with the properly, but even after years of Ruby and Python, it came as a surprise to me.  I think just a quick one-liner to explain that was your goal, or to have something documenting the multitype support for the method definition.

zeroSteiner: I added some YARD docs for the parameter that call out it accepts either a String or `Share::Provider::Base` instance. Let me know if you think I can make it more clear.

bwatters-r7:
# Release Notes
This PR adds the `remove_share` method to the server to support multiple smb shares.

PR #262, opened 2024-02-27
Add desired delete permissions when opening directory

dwelch-r7:
Drafted while we debug some issues with psexec

Pins the bindata version to 2.4.15 as 2.5.0 introduces significant changes that aren't currently compatible with ruby smb

Also aligns the permissions set when openign a directory with those of opening a file

Used by https://github.com/rapid7/metasploit-framework/pull/18895

### DIFF
--- a/lib/ruby_smb/smb2/tree.rb
+++ b/lib/ruby_smb/smb2/tree.rb
@@ -147,10 +147,11 @@ module RubySMB
       # @raise [RubySMB::Error::InvalidPacket] if the response is not a CreateResponse packet
       def open_directory(directory: nil, disposition: RubySMB::Dispositions::FILE_OPEN,
                          impersonation: RubySMB::ImpersonationLevels::SEC_IMPERSONATE,
-                         read: true, write: false, delete: false)
+                         read: true, write: false, delete: false, desired_delete: false)
 
         create_request  = open_directory_packet(directory: directory, disposition: disposition,
-                                                impersonation: impersonation, read: read, write: write, delete: delete)
+                                                impersonation: impersonation, read: read, write: write, delete: delete,
+                                                desired_delete: desired_delete)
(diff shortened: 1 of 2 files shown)


lib/ruby_smb/smb2/tree.rb
```diff
@@ -186,11 +186,22 @@ def open_directory_packet(directory: nil, disposition: RubySMB::Dispositions::FI
         create_request.create_options.directory_file  = 1
         create_request.file_attributes.directory      = 1
         create_request.desired_access.list            = 1
-        create_request.share_access.read_access       = 1 if read
-        create_request.share_access.write_access      = 1 if write
-        create_request.share_access.delete_access     = 1 if delete
         create_request.create_disposition             = disposition
 
+        if read
```
adfoster-r7, line 191: I'll defer to @cdelafuente-r7 who might have more insights here on the ramifications of these changes 😄

lib/ruby_smb/smb2/tree.rb
```diff
@@ -186,11 +186,22 @@ def open_directory_packet(directory: nil, disposition: RubySMB::Dispositions::FI
...
+          create_request.share_access.read_access = 1
+          create_request.desired_access.read_attr = 1
+        end
+
+        if write
+          create_request.share_access.write_access   = 1
```
adfoster-r7, line 197: I'm guessing there's no equivalent `create_request.desired_access` change needed here?

adfoster-r7:
This _might_ have broken psexec if we merge these changes, let's hold off on it

dwelch-r7:
> This _might_ have broken psexec if we merge these changes, let's hold off on it

this wasn't the issue with psexec as it turns out, made changes anyway to be more intentional with which ppermissions the user requests

PR #263, opened 2024-03-15
Adds EOFError raise when a session is dead

cgranleese-r7:
> [!note]
This PR is in conjunction with a PR in [Metasploit-Framework](https://github.com/rapid7/metasploit-framework/pull/18960) . This PR will need to be landed before the framework PR.

This PR is to add a more precise error being raised when a session is dead. Currently in Metasploit-Framework if a session dies we would have had a `RubySMB::Error::EncryptionError` error being raised which wasn't accurate. This change will now raise an `RubySMB::Error::CommunicationError` when a session dies, which feels a lot closer in terms of accuracy.

### DIFF
--- a/lib/ruby_smb/client.rb
+++ b/lib/ruby_smb/client.rb
@@ -578,9 +578,9 @@ module RubySMB
         raw_response = dispatcher.recv_packet
       rescue RubySMB::Error::CommunicationError => e
         if encrypt
-          raise RubySMB::Error::EncryptionError, "Communication error with the "\
-            "remote host: #{e.message}. The server supports encryption but was "\
-            "not able to handle the encrypted request."
+          raise e, "Communication error with the "\
+            "remote host: #{e.message}. The server supports encryption and this error "\
+            "may have been caused by encryption issues, but not always."
         else
           raise e
         end
--- a/spec/lib/ruby_smb/client_spec.rb
+++ b/spec/lib/ruby_smb/client_spec.rb
@@ -471,9 +471,9 @@ RSpec.describe RubySMB::Client do
       it 'raises an EncryptionError exception if an error occurs while receiving the response' do
         allow(dispatcher).to receive(:recv_packet).and_raise(RubySMB::Error::CommunicationError)
         expect { client.recv_packet(encrypt: true) }.to raise_error(
-          RubySMB::Error::EncryptionError,
+          RubySMB::Error::CommunicationError,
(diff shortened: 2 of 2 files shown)


lib/ruby_smb/client.rb
```diff
@@ -575,6 +575,9 @@ def recv_packet(encrypt: false)
       begin
         raw_response = dispatcher.recv_packet
       rescue RubySMB::Error::CommunicationError => e
+        # If raw_response is nil the connection is dead
+        raise EOFError if raw_response.nil?
```
adfoster-r7, line 579: Generally third party libraries don't raise raw exceptions like this, normally they're encapsulated and namespaced by the gem in question

i.e. what happened here: 

https://github.com/rapid7/ruby_smb/blob/0611a3e5b745d1ae5e49cd3514ad7d3c30132811/lib/ruby_smb/dispatcher/socket.rb#L87-L88

I think this try/catch logic that you're trying to update is the real problem that needs to be re-evaluated

cdelafuente-r7 (approved):

lib/ruby_smb/client.rb
```diff
@@ -578,7 +578,7 @@ def recv_packet(encrypt: false)
         raw_response = dispatcher.recv_packet
       rescue RubySMB::Error::CommunicationError => e
         if encrypt
-          raise RubySMB::Error::EncryptionError, "Communication error with the "\
+          raise e, "Communication error with the "\
```
adfoster-r7, line 581: For the context that you plan to use this in; if the session is killed - will the message be shown to the user as an encryption error?

cgranleese-r7: Good point. Should I add logic above this to check if the response is nil and raise a specific message for that scenario?
```suggestion

     raise RubySMB::Error::CommunicationError, "An error occurred reading from the Socket #{e.message}" if raw_response.nil?

      if encrypt
          raise e, "Communication error with the "\
``````

```

adfoster-r7: I think long-term we should probably handle this better, potentially special casing this error if it's got a connection to a windows 8 target

For now, maybe we could just re-word the error message to be less clear-cut that the communication error was caused by  encryption issues

```
      if encrypt
          raise e, "Communication error with the "\
            "remote host: #{e.message}. The server supports encryption and this error may have been caused by encryption issues, but not always."
```

Or maybe we could have special case for the amount of encrypted messages sent already, and if it's more than 1 - we just don't need this message any more

PR #264, opened 2024-03-28
Handle SMB2 compound related requests

zeroSteiner:
This adds handling for compound SMB2 requests as defined here: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-smb2/46dd4182-62d3-4e30-9fe5-e2ec124edca1

I ran into this while working on an exploit that triggered a load over a UNC path from a Python app running on a Windows Server 2022 host. In this case, there was a create request followed by another request whose `#file_id` attribute was not populated leading the higher level read operation to fail.

### DIFF
--- a/lib/ruby_smb/server/server_client.rb
+++ b/lib/ruby_smb/server/server_client.rb
@@ -36,6 +36,7 @@ module RubySMB
 
         # session id => session instance
         @session_table = {}
+        @smb2_related_operations_state = {}
       end
 
       #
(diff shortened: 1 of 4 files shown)


cdelafuente-r7:
Thanks @zeroSteiner! It looks good to me. I tested using the `examples/virtual_file_server.rb` script and made sure compound requests are handled correctly. However, I was not able to reproduce the same use case as you described. I only observed Create requests chained with Find requests. That said, the `file_id` is correctly populated and the find operations properly works. I'll go ahead and land it. Thanks!

cdelafuente-r7:
# Release Notes
This adds handling for compound SMB2 requests. Now, chained requests are correctly following the protocol and attributes like `session_id`, `tree_id` and `file_id` are properly populated across these requests.

PR #265, opened 2024-04-02
Add GetKeySecurity and SetKeySecurity MS-RRP structures (DCERPC)

cdelafuente-r7:
This adds two new operations to the [[MS-RRP]: Windows Remote Registry Protocol](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rrp/0fa3191d-bb79-490a-81bd-54c2601b7a78) DCERPC implementation.

- [3.1.5.13 BaseRegGetKeySecurity (Opnum 12)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rrp/b0e1868c-f4fd-4b43-959f-c0f0cac3ee26)
- [3.1.5.21 BaseRegSetKeySecurity (Opnum 21)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rrp/da18856c-8a6d-4217-8e93-3625865e562c)

This also includes a fix when passing an empty string to DCERPC requests, via RPC unicode strings or standard strings. This was causing NDR fault errors.
Now, it ensures at least one character (the string terminator) is counted. This follows the NDR protocol (see https://pubs.opengroup.org/onlinepubs/9629399/chap14.htm#tagcjh_19_03_04): the NDR String must contain at least one element, the terminator.

As a side note, it makes me wonder if the non-null-terminated NDR structures are relevant. NDR strings seems to always include the terminator. We might want to remove these structures and keep their null-terminated variant (same name but ending with `z`): `NdrVarString`, `NdrVarWideString`, `NdrConfVarString` and `NdrConfVarWideString`. Thta being said, it is too much changes for this PR and should be part of another PR.

### Testing
This includes a new example script `examples/registry_key_security_descriptor.rb` and can be used to test (it is documented in the script itself).

### DIFF
--- a/lib/ruby_smb/dcerpc/winreg.rb
+++ b/lib/ruby_smb/dcerpc/winreg.rb
@@ -71,6 +71,8 @@ module RubySMB
 
       BUFFER_SIZE = 1024
 
+      RegValue = Struct.new(:type, :data)
+
       # Open the registry root key and return a handle for it. The key can be
       # either a long format (e.g. HKEY_LOCAL_MACHINE) or a short format
       # (e.g. HKLM)
@@ -132,11 +134,10 @@ module RubySMB
       # @param handle [Ndr::NdrContextHandle] the handle for the key
       # @param value_name [String] the name of the value
       # @param value_name [Boolean] also return the data type if set to true
-      # @return [String] the data of the value entry
-      # @return [Array] if `type` is true, an array containing the data type and the actual data of the value entry
+      # @return [RegValue] a RegValue struct containing the data type and the actual data of the value entry
       # @raise [RubySMB::Dcerpc::Error::InvalidPacket] if the response is not a QueryValueResponse packet
       # @raise [RubySMB::Dcerpc::Error::WinregError] if the response error status is not ERROR_SUCCESS
-      def query_value(handle, value_name, type: false)
+      def query_value(handle, value_name)
         query_value_request_packet = RubySMB::Dcerpc::Winreg::QueryValueRequest.new(hkey: handle, lp_value_name: value_name)
         query_value_request_packet.lp_type = 0
         query_value_request_packet.lpcb_data = 0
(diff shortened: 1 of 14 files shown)


lib/ruby_smb/dcerpc/winreg.rb
```diff
@@ -124,14 +127,16 @@ def open_key(handle, sub_key)
...
 
       # Retrieve the data associated with the named value of a specified
-      # registry open key.
+      # registry open key. This will also return the type if required.
       #
       # @param handle [Ndr::NdrContextHandle] the handle for the key
       # @param value_name [String] the name of the value
+      # @param value_name [Boolean] also return the data type if set to true
       # @return [String] the data of the value entry
+      # @return [Array] if `type` is true, an array containing the data type and the actual data of the value entry
```
smcintyre-r7, line 136: This feels pretty off. Seems like we should have a new `#query_value` method that returns a hash containing the type and data so we could add new fields in the future and keep this one to call the new one and just return the data. This would ensure the return type is consistent rather than being dependent on the arguments the user is supplying.

Do you have any ideas how many places use this method that would need to be refactored if we just changed it directly? If we did that, we could return a struct with `type` and `data` attributes.

smcintyre-r7 (changes requested):

lib/ruby_smb/dcerpc/winreg.rb
cdelafuente-r7: I agree, it was a quick workaround to solve an issue, but we can think about something more robust.

I quickly checked in RubySMB and Metasploit and `#query_value` is not used that much. I didn't want to make the change directly in this method to avoid breaking changes. This would require a major version bump and might not be what we want.

That said, I'm fine with doing it as long as it is ok to release a new major release of RubySMB.

Otherwise, I also think the other option is good as a temporary change. We can still add this new `#query_value` method, as you suggested, with a warning saying it will disappear in the next major release, once we change the existing `#query_value` method.

cdelafuente-r7: After our discussion on slack, we decided to go ahead and change the `#query_value` method and return a struct with `type` and `data` attributes. We won't do a major version bump for this, even if it changes slightly the return value type. This method is only used in RubySMB and in one location in Metasploit.

smcintyre-r7 (approved):

PR #267, opened 2024-04-24
Add OpenPolicy and QueryInfoPolicy LSARPC implementations

sjanusz-r7:
This PR adds in LSARPC OpenPolicy, OpenPolicy2, QueryInfoPolicy, QueryInfoPolicy2, LookupSids and CloseHandle calls, and tests.

### DIFF
--- a/lib/ruby_smb/dcerpc/lsarpc.rb
+++ b/lib/ruby_smb/dcerpc/lsarpc.rb
@@ -1,10 +1,642 @@
+require 'ruby_smb/dcerpc/ndr'
+require 'ruby_smb/dcerpc/rrp_rpc_unicode_string'
+require 'ruby_smb/dcerpc/samr/rpc_sid'
+require 'ruby_smb/dcerpc/uuid'
+
 module RubySMB
   module Dcerpc
     module Lsarpc
+
       # see: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dssp/6f843846-2494-4d49-b715-2f181317dd34
-      UUID = 'c681d488-d850-11d0-8c52-00c04fd90f7e'.freeze
-      VER_MAJOR = 1
+      UUID = '12345778-1234-abcd-ef00-0123456789ab'.freeze
+      VER_MAJOR = 0
       VER_MINOR = 0
+
+      # OPNUMS
+      LSAR_CLOSE_HANDLE               = 0
+      LSAR_OPEN_POLICY                = 6
+      LSAR_QUERY_INFORMATION_POLICY   = 7
+      LSAR_LOOKUP_SIDS                = 15
+      LSAR_OPEN_POLICY2               = 44
+      LSAR_QUERY_INFORMATION_POLICY2  = 46
+
+      ################
+      # ACCESS_MASK Values
+
+      # [2.2.1.1 ACCESS_MASK](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lsad/7aeb7f17-0a6e-4f04-ac7e-7b1363cf9ecf)
+      # [2.4.3 ACCESS_MASK](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-dtyp/7a53f60e-e730-4dfe-bbe9-b21b62eb790b)
+      DELETE                 = 0x00010000
+      READ_CONTROL           = 0x00020000
+      WRITE_DACL             = 0x00040000
+      WRITE_OWNER            = 0x00040000
+      SYNCHRONIZE            = 0x00100000
+      ACCESS_SYSTEM_SECURITY = 0x01000000
+      MAXIMUM_ALLOWED        = 0x02000000
+      GENERIC_ALL            = 0x10000000
+      GENERIC_EXECUTE        = 0x20000000
+      GENERIC_WRITE          = 0x40000000
+      GENERIC_READ           = 0x80000000
+
+      ################
+      # SECURITY_DESCRIPTOR_CONTROL
+
+      # [SECURITY_DESCRIPTOR_CONTROL](https://learn.microsoft.com/en-us/windows/win32/secauthz/security-descriptor-control)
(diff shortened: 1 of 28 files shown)


lib/ruby_smb/dcerpc/lsarpc/lsar_lookup_sids_request.rb
```diff
@@ -0,0 +1,26 @@
...
+    module Lsarpc
+
+      # This class represents a LsarLookupSids Request Packet as defined in
+      # [3.1.4.11 LsarLookupSids (Opnum 15)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lsat/eb7ac899-e697-4883-93de-1e60c7720c02)
+      class LsarLookupSidsRequest < BinData::Record
+        attr_reader :opnum
+
+        endian :little
+
+        lsapr_handle :policy_handle
```
cgranleese-r7, line 12: Get the definitions here aligned with the same number of spaces

lib/ruby_smb/dcerpc/lsarpc.rb
```diff
@@ -1,11 +1,644 @@
...
+      class SecurityQualityOfServicePtr < SecurityQualityOfService
+        extend Ndr::PointerClassPlugin
+      end
+
+      # [2.2.2.4 LSAPR_OBJECT_ATTRIBUTES](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lsad/ad9e183d-6474-4641-a6d9-d3796d2d604b)
+      class LsaprObjectAttributes < Ndr::NdrStruct
+        default_parameters byte_align: 4
+        endian :little
+
+        ndr_uint32                      :len
```
smcintyre-r7, line 211: So I'm looking at [`#open_policy_2`](https://github.com/rapid7/metasploit-framework/pull/19170/files#diff-32c8b2752caef945fe888b76e002c9ce16113645fe0b720992b9f51ed214e988R52) and I'm wondering if it'd be possible to have this `len` field be calculated by default so the caller doesn't have to set it in most cases. Same thing goes for the `len` field of `SecurityQualityOfService`.

lib/ruby_smb/dcerpc/lsarpc.rb
```diff
@@ -1,11 +1,644 @@
...
+          lsar_response = LsarLookupSidsResponse.read(response)
+        rescue IOError
+          raise RubySMB::Dcerpc::Error::InvalidPacket, 'Error reading LsarLookupSidsResponse'
+        end
+        unless lsar_response.error_status == WindowsError::NTStatus::STATUS_SUCCESS
+          raise RubySMB::Dcerpc::Error::LsarpcError,
+                "Error returned while looking up SID: "\
+                  "#{WindowsError::NTStatus.find_by_retval(lsar_response.error_status.value).join(',')}"
+        end
+        lsar_response.translated_names
```
smcintyre-r7, line 639: Instead of returning the `LsaprTranslatedNames` instance, it might be nice to just return an array of strings of the translated names so the caller doesn't need to deal with the structure.

lib/ruby_smb/dcerpc/lsarpc.rb
```diff
@@ -1,11 +1,644 @@
...
+        end
+        unless lsar_response.error_status == WindowsError::NTStatus::STATUS_SUCCESS
+          raise RubySMB::Dcerpc::Error::LsarpcError,
+                "Error returned while closing policy handle: "\
+                  "#{WindowsError::NTStatus.find_by_retval(lsar_response.error_status.value).join(',')}"
+        end
+        lsar_response.policy_handle
+      end
+
+      def lsar_lookup_sids(policy_handle:, sid_enum_buffer:, lookup_level:)
```
smcintyre-r7, line 620: I think I commented on this else where but it would be easier to sue if the caller didn't need to deal constructing the sid_enum_buffer themselves but instead could pass an array of SID strings and this function would translate that into the necessary object instance for BinData.

smcintyre-r7 (commented):
Just a couple of general comments:

* It's not necessary to initialize values to `0` and `nil` in many cases. Those values will be used by default if the keyword isn't passed to the `BinData::Record`'s constructor. You can also change the default value by setting it to something static or [even a function](https://github.com/dmendel/bindata/wiki/Records#dependencies-between-fields) which is particularly useful for length fields. Using `default_value` instead of `value` also means that it can be changed.

* I called out the names ending in `_2` and from what I've seen the precedence within RubySMB is almost entirely to prefer omitting the underscore. For example `ruby_smb/dcerpc/samr/samr_set_information_user2` and `ruby_smb/dcerpc/netlogon/netr_server_password_set2_request.rb`. I'm pretty sure the pattern would extrapolate to constants and function names as well.

smcintyre-r7 (approved):
I think we're all set here now. The changes are looking good when tested with rapid7/metasploit-framework#19170.

```
metasploit-framework (S:0 J:0) auxiliary(scanner/smb/smb_lookupsid) > run MinRID=570 MaxRID=1105 VERBOSE=true

[*] 3.22.14.197:445 - Connecting to Local Security Authority Remote Protocol
[*] 3.22.14.197:445 - Binding to \lsarpc...
[+] 3.22.14.197:445 - Bound to \lsarpc
[*] 3.22.14.197:445 - PIPE(lsarpc) LOCAL(labs1collabu0 - S-1-5-21-795503-3050334394-3644400624) DOMAIN(labs1collabu0 - S-1-5-21-795503-3050334394-3644400624)
[*] Trying RID 1105 / 1105
SMB Lookup SIDs Output
======================

    Type   Name                                     RID
    ----   ----                                     ---
    Alias  Allowed RODC Password Replication Group  571
    Alias  Denied RODC Password Replication Group   572
    User   SRV-ADDS01$                              1000
    Alias  DnsAdmins                                1101
    Group  DnsUpdateProxy                           1102
    User   mmoose                                   1103
    User   idr_service_account                      1104
    Group  Austin Server Admins Group               1105

[*] 3.22.14.197: - Scanned 1 of 1 hosts (100% complete)
[*] Auxiliary module execution completed
metasploit-framework (S:0 J:0) auxiliary(scanner/smb/smb_lookupsid) >
````

PR #269, opened 2024-05-31
Adding dsrgetdcnameex2 implementation

sud0Ru:
I need this DCERPC [call](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/fb8e1146-a045-4c31-98d1-c68507ad5620) for my Metasploit module. I successfully implemented the request, but when it came to implementing the response, I always encountered problems with fields pointing to incorrect data. For my module, it's sufficient to implement only the request. However, with some assistance from you, perhaps we can also implement the response.
Here is the response's code which should be in file called ```dsr_get_dc_name_ex2_resoponse.rb```:
```require 'ruby_smb/dcerpc/ndr'

module RubySMB
  module Dcerpc
    module Netlogon

      # [3.5.4.3.1 DsrGetDCNameEx2 (Opnum 34)](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/fb8e1146-a045-4c31-98d1-c68507ad5620)
      class DsrGetDCNameEx2Response < BinData::Record
        attr_reader :opnum

        endian :little

        pdomain_controller_info_w :domain_info
        ndr_uint32             :error_status

        def initialize_instance
          super
          @opnum =  DSR_GET_DC_NAME_EX2
        end
      end
    end
  end
end
```
Inside ```netlogin.rb``` there is the implementation for this [struct ](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/9b85a7a4-8d34-4b9e-9500-bf8644ebfc06):
```
      class DomainControllerInfoW < Ndr::NdrStruct
        default_parameter byte_align: 4
        endian :little

        ndr_wide_stringz_ptr :dc_name
        ndr_wide_stringz_ptr :dc_address
        ndr_uint32 :dc_address_type
        uuid_ptr :domain_guid
        ndr_wide_stringz_ptr :domain_name
        ndr_wide_stringz_ptr :domain_forrest
        ndr_uint32 :flags
        ndr_wide_stringz_ptr :dc_site_name
        ndr_wide_stringz_ptr :client_site_name
      end

      class PdomainControllerInfoW < DomainControllerInfoW
        extend Ndr::PointerClassPlugin
      end
```
If I want to access for example `dc_address` I will get the data that related to `dc_name`
for `dc_name` I will get some non-printable bytes which located some where in the beginning of the response

### DIFF
--- a/lib/ruby_smb/dcerpc/netlogon.rb
+++ b/lib/ruby_smb/dcerpc/netlogon.rb
@@ -11,6 +11,7 @@ module RubySMB
       NETR_SERVER_REQ_CHALLENGE = 4
       NETR_SERVER_AUTHENTICATE3 = 26
       NETR_SERVER_PASSWORD_SET2 = 30
+      DSR_GET_DC_NAME_EX2 = 34
 
       # see: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nrpc/3b224201-b531-43e2-8c79-b61f6dea8640
       class LogonsrvHandle < Ndr::NdrWideStringzPtr; end
@@ -65,6 +66,7 @@ module RubySMB
       require 'ruby_smb/dcerpc/netlogon/netr_server_password_set2_response'
       require 'ruby_smb/dcerpc/netlogon/netr_server_req_challenge_request'
       require 'ruby_smb/dcerpc/netlogon/netr_server_req_challenge_response'
+      require 'ruby_smb/dcerpc/netlogon/dsr_get_dc_name_ex2_request'
 
       # Calculate the netlogon session key from the provided shared secret and
       # challenges. The shared secret is an NTLM hash.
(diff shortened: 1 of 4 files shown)


smcintyre-r7:
I sent a PR to your branch for you to test and review with the response class implemented in sud0Ru#1. The changes you have here all look good. I just tweaked the whitespace so things are alinged and changed `DC` to Dc` to match the docs.

Once you merge my changes, I'll merge this and get the changes released in a gem bump. Once that's done I'll walk you through updating your Metasploit module.

Thanks for taking the time to put this in RubySMB!

sud0Ru:
Thank you for the review. I will check the PR.

From a quick look at your commits, it seems the issue in my code was using `uuid_ptr` instead of `uuid` in the `DomainControllerInfoWstruct` struct, correct? :)

smcintyre-r7:
> From a quick look at your commits, it seems the issue in my code was using uuid_ptr instead of uuid in the DomainControllerInfoWstruct struct, correct? :)

I think that's correct, it sounds about right.

I've gone ahead and merged this. I'll get it released in a new version shortly which will be 3.3.9.

PR #278, opened 2024-11-14
Add the ability to query group memberships

smashery:
Added two SAMR messages, enabling querying group membership.

Supporting https://github.com/rapid7/metasploit-framework/pull/19643

### DIFF
--- a/lib/ruby_smb/dcerpc/samr/samr_get_members_in_group_response.rb
+++ b/lib/ruby_smb/dcerpc/samr/samr_get_members_in_group_response.rb
@@ -25,7 +25,7 @@ module RubySMB
 
         def initialize_instance
           super
-          @opnum = SAMR_GET_GROUPS_FOR_USER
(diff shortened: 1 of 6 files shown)


lib/ruby_smb/dcerpc/samr/samr_get_members_in_group_response.rb
```diff
@@ -0,0 +1,34 @@
...
+        attr_reader :opnum
+
+        endian :little
+
+        psampr_get_members_buffer :members
+        ndr_uint32               :error_status
+
+        def initialize_instance
+          super
+          @opnum = SAMR_GET_GROUPS_FOR_USER
```
smcintyre-r7, line 28: I think this is supposed to be `SAMR_GET_MEMBERS_IN_GROUP`
```suggestion
          @opnum = SAMR_GET_MEMBERS_IN_GROUP
```

smcintyre-r7 (changes requested):

PR #280, opened 2024-11-21
Fix two fields in UserProperties

zeroSteiner:
This fixes the definition of the `PropertyCount` and `UserProperties` fields that can be omitted. Based on the [docs](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-samr/8263e7ab-aba9-43d2-8a36-3a9cb2dd3dad).

> When there are zero USER_PROPERTY elements in the UserProperties field, this field MUST be omitted; the resultant USER_PROPERTIES structure has a constant size of 0x6F bytes.

This allows the structure to be read correctly even when there are not user properties. Allowing the structure to be read fixes the IOError that's raised when there are no kerberos keys as identified in rapid7/metasploit-framework#19665.

### DIFF
--- a/lib/ruby_smb/dcerpc/samr.rb
+++ b/lib/ruby_smb/dcerpc/samr.rb
@@ -443,8 +443,8 @@ module RubySMB
         uint16 :reserved3
         string :reserved4, length: 96
         uint16 :property_signature, initial_value: 0x50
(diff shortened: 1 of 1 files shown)


smashery:
Tested and all works well! Much better solution - thanks!

PR #294, opened 2026-03-14
Add Windows 95/98/ME SMB1 support

Z6543:
Fixes #293

  ## Summary

  - Add SMB1 protocol support for Windows 95/98/ME and other pre-NT legacy hosts that use port 139, non-extended security, share-level authentication, and RAP-based share enumeration
  - Fix `TreeConnectRequest` password field serialization bug that corrupted packets when callers set `password_length` without a trailing null
  - Add NetBIOS name resolution fallback when the `*SMBSERVER` wildcard is rejected

  ## Details

  ### Windows 95/98/ME SMB1 support (`d0d2607`)

  Windows 9x hosts differ from NT-based systems in several ways that RubySMB did not handle:

  - **Non-extended security negotiate**: parse the legacy negotiate response (no `extended_security` flag, no NTLMSSP)
  - **Legacy LM/NTLM session setup**: use `SessionSetupLegacyRequest`/`SessionSetupLegacyResponse` instead of NTLMSSP blobs
  - **Minimal TreeConnectResponse**: Win95 may omit the `native_file_system` field; the response DataBlock now handles short reads
  - **FindInfoStandard**: add the `SMB_FIND_FILE_STANDARD` information level for TRANS2 directory listings
  - **SMB1::Tree fallbacks**: handle `ERRDOS/ERRnomem` from TRANS2 by falling back to core protocol commands

  ### SMB_COM_OPEN_ANDX, share-level auth, and RAP (`10ed8be`)

  - **`net_share_enum_rap`**: enumerate shares via RAP (Remote Administration Protocol) over `\PIPE\LANMAN`, the only method Windows 9x supports (DCERPC/SRVSVC fails with `DBG_CONTINUE`)
  - **`smb1_tree_connect` password support**: accept a `password:` keyword for share-level authentication where passwords are per-share, not per-user
  - **`SMB_COM_OPEN_ANDX`**: add request/response packets and `Tree#open_file`/`Tree#read_file`/`Tree#write_file` for Win95 hosts that lack NT_CREATE_ANDX

  ### Fix TreeConnectRequest password field (`541321c`)

  The password field was declared as `BinData::Stringz`, which always appends a null terminator when serializing, regardless of the `length` parameter. With `password_length=N`, the field wrote N+1 bytes, shifting the share path and corrupting the packet.
  Changed to `BinData::String` which respects the length exactly. Existing callers are unaffected since `smb1_tree_connect` already appends `"\x00"` explicitly.

[...]

Issue #19, opened 2015-04-15
Query Access Info


jlee-r7:
https://msdn.microsoft.com/en-us/library/cc232116.aspx

Also [MSDN 2.2.37 SMB2 QUERY_INFO Request](https://msdn.microsoft.com/en-us/library/cc246557.aspx)

Issue #16, opened 2015-04-15
Query Standard Info


jlee-r7:
https://msdn.microsoft.com/en-us/library/cc232088.aspx

Also [MSDN 2.2.37 SMB2 QUERY_INFO Request](https://msdn.microsoft.com/en-us/library/cc246557.aspx)

jlee-r7:
Looks like all the information obtained from this request is actually available in the `CreateResponse` as well.

PR #296, opened 2026-04-23
Add NetBIOS name resolution fallback for session_request

Z6543:
## Summary

Follow-up to #294 addressing review comments #16, #17, #18, #19 and #20 from @smcintyre-r7. Splits the NetBIOS auto-discovery functionality that was removed from #294 into a dedicated reviewable unit.

On `CALLED_NAME_NOT_PRESENT` (NBSS error 0x82), `Client#session_request` raises `RubySMB::Error::NetBiosSessionService` carrying the numeric `error_code` rather than retrying internally. The caller resolves the server's real NetBIOS name with the pure-Ruby `RubySMB::Nbss::NodeStatus.file_server_name`, reconnects, and retries `session_request` with that name. Reconnect is left to the caller so the replacement socket is routed the same way as the original (e.g. through a Metasploit pivot).

## Relationship to #294

This branch contains the commits from #294 plus the NetBIOS-specific additions on top. Review is easiest after #294 merges — at that point the diff here will be only the NetBIOS work. If you'd prefer, I can rebase onto `master` directly once #294 lands.

## What's in the diff

- `RubySMB::Nbss::NodeStatus.query(host)` / `.file_server_name(host)` — public pure-Ruby `nmblookup -A` equivalent, built on new `Nbss::NodeStatusRequest` and `Nbss::NodeStatusResponse` BinData structures. No external binaries.
- Caller-owned UDP socket — `NodeStatus.query` takes a UDP socket the caller creates, binds, and closes (stdlib `UDPSocket` or `Rex::Socket::Udp`). The library creates no sockets and performs no reconnect, so the caller routes it however it needs (e.g. through a Metasploit pivot).
- `RubySMB::Error::NetBiosSessionService#error_code` — carries the numeric NBSS error code; the named constants live in `RubySMB::Nbss` (`CALLED_NAME_NOT_PRESENT` = 0x82, etc). Callers match on the number, not a message substring.

## Addresses PR #294 review comments

- **#16, #18** — the library no longer creates sockets; `NodeStatus.query` accepts a caller-owned UDP socket, and reconnect on retry is the caller's responsibility.
- **#17** — `netbios_lookup_nmblookup` shell-out is gone; everything is Ruby.
- **#19** — `lib/ruby_smb/nbss/node_status_request.rb` and `node_status_response.rb` follow the existing NBSS pattern; the UDP lookup is ~10 lines of BinData glue.
- **#20** — auto-discovery moved out of #294 (commit 09c848c there) into this dedicated branch.

## Test plan

- [x] `bundle exec rspec` — 12,300 examples, 0 failures (+14 new specs: `node_status_request_spec.rb`, `node_status_response_spec.rb`, `node_status_spec.rb`, plus NBSS `error_code` propagation coverage in `client_spec.rb`).
- [x] End-to-end against a live Windows 95 target (via Metasploit's cve_2000_0979 module): session_request retry fires, NodeStatus resolves the real name, tree-connect proceeds.
- [x] Verified pivoted path works when the MSF caller supplies a `Rex::Socket::Udp.create` factory.
- [ ] Reviewer concurrence on the caller-owned socket and caller-owned reconnect approach.


smcintyre-r7:
Since you split this out there's conflicts. It looks like there was some shared commits in this branch and in #294 which has been landed. Can you please rebase this on `master` so it'll merge cleanly? It should include those changes and then I can get to reviewing and testing this. It'll also make review easier because it won't keep changes that have already been merged.

Z6543:
The Copilot comments are no longer valid for this PR after the rebase. 

There is one important "challenge" with this PR. Metasploit has to be able to listen on port 137. Even though the smb client sends the UDP packet from a random high port to port 137, the smb server responds to port 137 from port 137.

But on the other hand, this PR is only needed if the Metasploit user does not provide the SMBSERVER parameter and it is left on default. 

Here is the output of nmblookup -A 192.168.124.178
```
Request:
12:23:09.064726 IP 192.168.124.69.49411 > 192.168.124.178.137: UDP, length 50
        0x0000:  4500 004e a7fc 4000 4011 185a c0a8 7c45  E..N..@.@..Z..|E
        0x0010:  c0a8 7cb2 c103 0089 003a 7a94 1db8 0000  ..|......:z.....
        0x0020:  0001 0000 0000 0000 2043 4b41 4141 4141  .........CKAAAAA
        0x0030:  4141 4141 4141 4141 4141 4141 4141 4141  AAAAAAAAAAAAAAAA
        0x0040:  4141 4141 4141 4141 4100 0021 0001       AAAAAAAAA..!..

Response, see ports: 
12:23:09.071411 IP 192.168.124.178.137 > 192.168.124.69.137: UDP, length 229
        0x0000:  4500 0101 9103 0000 2011 8ea0 c0a8 7cb2  E.............|.
        0x0010:  c0a8 7c45 0089 0089 00ed d867 1db8 8400  ..|E.......g....
        0x0020:  0000 0001 0000 0000 2043 4b41 4141 4141  .........CKAAAAA
        0x0030:  4141 4141 4141 4141 4141 4141 4141 4141  AAAAAAAAAAAAAAAA
        0x0040:  4141 4141 4141 4141 4100 0021 0001 0000  AAAAAAAAA..!....
        0x0050:  0000 009b 0644 4546 4155 4c54 2020 2020  .....DEFAULT....
        0x0060:  2020 2020 0044 0057 4f52 4b47 524f 5550  .....D.WORKGROUP
....                                  
```

lib/ruby_smb/nbss/node_status_response.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      endian :big
+
+      string :netbios_name, label: 'NetBIOS Name', length: 15
+      uint8  :suffix,       label: 'Suffix'
+      uint16 :name_flags,   label: 'Name Flags'
+
+      def group?
+        (name_flags & GROUP_BIT) != 0
+      end
```
smcintyre-r7: That flags should really be defined as a bit field using [Bindata](https://github.com/dmendel/bindata/wiki/PrimitiveTypes#bit-based-integers). This becomes important if an instance is returned to another namespace where the GROUP_BIT may not already be defined such as a Metasploit module

That would allow the Metasploit module to do `node_status_name.name_flags.group = 1` instead of:

```ruby
require 'ruby_smb' # first it has to be require'ed so RubySMB is even defined
...
# next you have to do the bit operation |=, presenting an opportunity to accidentally omit the | and thus introduce a subtle bug
node_status_name.name_flags |= RubySMB::Nbss::NodeStatusName::GROUP_BIT
```

lib/ruby_smb/nbss/node_status_request.rb
```diff
@@ -0,0 +1,29 @@
...
+      # NBSTAT question type, RFC 1002 4.2.1.3.
+      QUESTION_TYPE_NBSTAT = 0x0021
+      # Internet class.
+      QUESTION_CLASS_IN    = 0x0001
+
+      endian :big
+
+      # 12-byte NBNS header (RFC 1002 4.2.1.1 and 4.2.1.2).
+      uint16 :transaction_id,  label: 'Transaction ID'
+      uint16 :flags,           label: 'Flags',     initial_value: 0x0000
```
smcintyre-r7: Same comment here regarding the `flags` field. If they're never used because no flag is ever set, you could cut corners by just skipping this field and defining it as a padding. If it's used though because one or more flag is set, then it needs to be fully defined.

lib/ruby_smb/nbss/node_status_response.rb
```diff
@@ -0,0 +1,66 @@
...
+    # [RFC 1002 4.2.18](https://tools.ietf.org/html/rfc1002#section-4.2.18).
+    # Received over UDP from port 137 in reply to a {NodeStatusRequest}.
+    # Does not decode the trailing STATISTICS field; callers only need the
+    # name table.
+    class NodeStatusResponse < BinData::Record
+      endian :big
+
+      # 12-byte NBNS header.
+      uint16 :transaction_id, label: 'Transaction ID'
+      uint16 :flags,          label: 'Flags'
```
smcintyre-r7: Same comment regarding flags.

lib/ruby_smb/client.rb
```diff
@@ -667,20 +688,72 @@ def wipe_state!
...
+    # auto-discovery in {#session_request}.
+    def default_called_name?(name)
+      name.nil? || name.to_s.strip.empty? || name.to_s.strip.upcase == '*SMBSERVER'
+    end
+
+    # Sends a single NetBIOS Session Request and reads the response.
+    #
+    # @param name [String] the NetBIOS name to request
+    # @return [TrueClass] if session request is granted
+    def send_session_request(name)
```
smcintyre-r7: This name doesn't make sense if it's both sending and receiving. At this point it looks like this should be back in the body of `#session_request`. The body then that handles `RubySMB::Error::NetBiosSessionService` could be it's own method. The point is though that if this method sends and receives as the docs state, then it shouldn't use the `send` as the verb prefix on the subject (session_request).

smcintyre-r7:
Feel free to wait to address the copilot comments. I got about half way through the review when I noticed: https://github.com/rapid7/ruby_smb/pull/296/changes#diff-aaf7fad50c553ee60d508c1a582817c6a9534c40e591b6441c5449a7708dc73dR194-R224

I understand why those work arounds are necessary, but I'd prefer to fix the root of the problem. That's shouldn't be your responsibility though so I've opened rapid7/rex-socket#82 and rapid7/rex-socket#83 to fix the API discrepancies. I'll get back to reviewing this once those have been landed and we can get a clean solution in that doesn't need to work around those issues.

Sit tight and I'll let you know when I can get this moving again. At that time you'll be able to drop those work arounds and just use the socket and it won't matter if it's a `UDPSocket` or a `Rex::Socket::Udp`.

smcintyre-r7:
Those Rex::Socket issues have been resolved. If you want to take a look at the PR I sent you that removes the workarounds and socket factory pattern, we can get this moving forward again.

https://github.com/Z6543/ruby_smb/pull/3

Z6543:
Great, thank you, PR accepted

smcintyre-r7:
I sent one more PR that addresses the last of my comments which I found I never posted from May.

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,95 @@
...
+          next unless IO.select([udp_socket], nil, nil, timeout)
+
+          data, = udp_socket.recvfrom(4096)
+          next if data.nil? || data.empty?
+
+          response = NodeStatusResponse.read(data)
+          return entries_from(response)
+        end
+        nil
+      rescue IOError, EOFError
```
bwatters-r7: We may also want to rescue `SystemCallError` to catch `ECONNREFUSED` and `EHOSTUNREACH`

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,95 @@
...
+        request = NodeStatusRequest.new(transaction_id: rand(0xFFFF))
+        request.question_name.set('*'.ljust(16, "\x00"))
+        bytes = request.to_binary_s
+
+        retries.times do
+          udp_socket.send(bytes, 0, host, port)
+          next unless IO.select([udp_socket], nil, nil, timeout)
+
+          data, = udp_socket.recvfrom(4096)
+          next if data.nil? || data.empty?
```
bwatters-r7, line 59: Should we also ignore the message if the transaction ID or sender does not match?  We're just reading the udp socket _hoping_ it is the response we expect from the source we expect.  This could be a security issue.

spec/lib/ruby_smb/client_spec.rb
```diff
@@ -742,6 +742,20 @@
...
         expect { client.session_request }.to raise_error(RubySMB::Error::InvalidPacket)
       end
+
+      describe 'NetBiosSessionService error propagates the error_code' do
+        it 'attaches the numeric NBSS error code to the raised exception' do
+          negative = RubySMB::Nbss::NegativeSessionResponse.new
+          negative.session_header.session_packet_type = RubySMB::Nbss::NEGATIVE_SESSION_RESPONSE
+          negative.error_code = RubySMB::Nbss::NegativeSessionResponse::NOT_LISTENING_ON_CALLED_NAME
+          allow(dispatcher).to receive(:recv_packet).and_return(negative.to_binary_s)
+          begin
```
bwatters-r7: Is this the right way to do this?  Should it be `expect negative.to raise_error()` or something like that?

lib/ruby_smb/nbss/node_status_response.rb
```diff
@@ -0,0 +1,81 @@
...
+      ACTIVE_BIT = 0x0400  # 1 = name registered
+
+      endian :big
+
+      string :netbios_name, label: 'NetBIOS Name', length: 15
+      uint8  :suffix,       label: 'Suffix'
+      node_status_name_flags :name_flags, label: 'Name Flags'
+
+      def group?
+        name_flags[:group].to_i == 1
```
bwatters-r7: Should this be using the literals above?
You've defined GROUP_BIT above, but then use something else?

lib/ruby_smb/nbss/node_status.rb
Z6543: Good catch — added `SystemCallError` to the rescue so `ECONNREFUSED` (ICMP port-unreachable) and `EHOSTUNREACH` return `nil` instead of propagating. (fa57a06)

lib/ruby_smb/nbss/node_status.rb
Z6543: Agreed, real spoofing risk on an unconnected UDP socket. `query` now drops any datagram whose source address ≠ the queried host or whose transaction ID ≠ the one we sent, and keeps waiting within the timeout window. Added specs for both rejection paths. (fa57a06)

spec/lib/ruby_smb/client_spec.rb
Z6543: Right — the `rescue` block let a non-raising path pass silently. Switched to `expect { … }.to raise_error(RubySMB::Error::NetBiosSessionService) do |e| … end` so a missing raise now fails the example. (fa57a06)

lib/ruby_smb/nbss/node_status_response.rb
Z6543: Fixed — `group?`/`active?` now mask the raw NAME_FLAGS word with `GROUP_BIT`/`ACTIVE_BIT` instead of reaching into the bitfield, so the constants are the source of truth. Kept the `NodeStatusNameFlags` record since the parse specs assert its named fields. (fa57a06)

bwatters-r7:
@cdelafuente-r7, did you want to be assigned to this?

lib/ruby_smb/nbss/name_service_header_flags.rb
```diff
@@ -0,0 +1,18 @@
...
+
+      bit1 :response,             label: 'Response',             initial_value: 0
+      bit4 :opcode,               label: 'Opcode',               initial_value: 0
+      bit1 :authoritative_answer, label: 'Authoritative Answer', initial_value: 0
+      bit1 :truncated,            label: 'Truncated',            initial_value: 0
+      bit1 :recursion_desired,    label: 'Recursion Desired',    initial_value: 0
+      bit1 :recursion_available,  label: 'Recursion Available',  initial_value: 0
+      bit2 :reserved,             label: 'Reserved',             initial_value: 0
+      bit1 :broadcast,            label: 'Broadcast',            initial_value: 0
+      bit4 :rcode,                label: 'Result Code',          initial_value: 0
```
cdelafuente-r7, line 15: The layout is correct but the RFC defines other grouping: `OPCODE`, `NM_FLAGS`, and `RCODE` fields. In my understanding, `NameServiceHeaderFlags` should only be `NM_FLAGS`, if we want to follow the RFC.

lib/ruby_smb/nbss/node_status_request.rb
```diff
@@ -0,0 +1,31 @@
...
+      # NBSTAT question type, RFC 1002 4.2.1.3.
+      QUESTION_TYPE_NBSTAT = 0x0021
+      # Internet class.
+      QUESTION_CLASS_IN    = 0x0001
+
+      endian :big
+
+      # 12-byte NBNS header (RFC 1002 4.2.1.1 and 4.2.1.2).
+      uint16                    :transaction_id, label: 'Transaction ID'
+      name_service_header_flags :flags,          label: 'Flags'
```
cdelafuente-r7, line 18: Following up on my comment about the header flags, I'm wondering if sticking to the RFC would be a good idea:

```suggestion
      name_service_opcode       :opcode,         label: 'Packet type code'
      name_service_header_flags :nm_flags,       label: 'Flags'
      name_service_result_code  :rcode,          label: 'Result codes'
```

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,101 @@
...
+
+          response = NodeStatusResponse.read(data)
+          # Reject replies whose transaction ID doesn't match our request.
+          next unless response.transaction_id.to_i == request.transaction_id.to_i
+
+          return entries_from(response)
+        end
+        nil
+      rescue IOError, EOFError, SystemCallError
+        nil
```
cdelafuente-r7, line 72: This rescue is located outside of the `retries` loop. As a result, any exceptions raised during one iteration will cancel all the remaining retries. For exemple, if a malformed datagram is received during the first try, `NodeStatusResponse.read` will raise `IOError` and the loop will be immediately exited. No retries will be performed.

I suggest to move this rescue inside the loop. Also, specs to test the retry mechanism when an error occurs should be added.

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,101 @@
...
+      # @param host [String] target IP address (unicast — no broadcast)
+      # @param port [Integer] destination UDP port (default 137)
+      # @param timeout [Numeric] per-attempt receive timeout in seconds
+      # @param retries [Integer] total number of attempts
+      # @param udp_socket [UDPSocket, Rex::Socket::Udp] caller-owned UDP socket.
+      #   The caller is responsible for binding and closing it.
+      # @return [Array<Entry>, nil] the name table, or nil on timeout/parse failure
+      def self.query(host, port: NBNS_PORT, timeout: DEFAULT_TIMEOUT,
+                     retries: DEFAULT_RETRIES, udp_socket:)
+        request = NodeStatusRequest.new(transaction_id: rand(0xFFFF))
```
cdelafuente-r7, line 50: I recommend using `SecureRandom.random_number` for improved security. Also, the range should be `0x10000` and not `0xFFFF` to include the full range of possible values `0-0xFFFF`.

```suggestion
        request = NodeStatusRequest.new(transaction_id: SecureRandom.random_number(0x10000))
```
Note that `require 'securerandom'` might be needed and the specs might need to be updated.

lib/ruby_smb/nbss/negative_session_response.rb
```diff
@@ -5,22 +5,29 @@ module Nbss
     # Representation of the NetBIOS Negative Session Service Response packet as defined in
     # [4.3.4 SESSION REQUEST PACKET](https://tools.ietf.org/html/rfc1002)
     class NegativeSessionResponse < BinData::Record
+      # NBSS error codes (RFC 1002 section 4.3.6)
+      NOT_LISTENING_ON_CALLED_NAME       = 0x80
+      NOT_LISTENING_FOR_CALLING_NAME     = 0x81
+      CALLED_NAME_NOT_PRESENT            = 0x82
+      CALLED_NAME_INSUFFICIENT_RESOURCES = 0x83
+      UNSPECIFIED_ERROR                  = 0x8F
```
cdelafuente-r7, line 13: By convention, and to make sure constants used externally are at the same location. I would suggest to move these error code constants to `lib/ruby_smb/nbss.rb`. this file already defines the Session Packet Type constants.

Note that the reference below will need to be updates:
```ruby
  def error_msg
    case error_code
    when RubySMB::Nbss::NOT_LISTENING_ON_CALLED_NAME
    ...
```

lib/ruby_smb/client.rb
```diff
@@ -679,8 +687,11 @@ def session_request(name = '*SMBSERVER')
...
         session_header = RubySMB::Nbss::SessionHeader.read(raw_response)
         if session_header.session_packet_type == RubySMB::Nbss::NEGATIVE_SESSION_RESPONSE
-          negative_session_response =  RubySMB::Nbss::NegativeSessionResponse.read(raw_response)
-          raise RubySMB::Error::NetBiosSessionService, "Session Request failed: #{negative_session_response.error_msg}"
+          negative_session_response = RubySMB::Nbss::NegativeSessionResponse.read(raw_response)
+          raise RubySMB::Error::NetBiosSessionService.new(
+            "Session Request failed: #{negative_session_response.error_msg}",
+            error_code: negative_session_response.error_code
+          )
         end
```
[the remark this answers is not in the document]
cdelafuente-r7: I believe this is a real issue. I cannot find where `RubySMB::Nbss::NodeStatus.file_server_name` is called to retry when `CALLED_NAME_NOT_PRESENT` is returned. The description of this PR states it should be solved. But, the method documentation above states that this should be the caller responsibility, which differs from this PR description. Please, can you elaborate?

Also, a specific test should be added in the specs to test the scenario where `CALLED_NAME_NOT_PRESENT` is returned.

lib/ruby_smb/nbss/node_status_response.rb
```diff
@@ -0,0 +1,89 @@
...
+      # RDATA begins here. NODE_NAME_ARRAY is preceded by an 8-bit count.
+      uint8  :num_names,        label: 'Number of Names'
+      array  :node_names, type: :node_status_name, initial_length: :num_names
+
+      # Returns the unique (non-group) file-server name (suffix 0x20) if one
+      # is present in the name table, else nil.
+      def file_server_name
+        entry = node_names.find { |n| n.suffix == 0x20 && n.unique? }
+        entry&.netbios_name&.to_s&.rstrip
+      end
```
cdelafuente-r7, line 86: It looks like `file_server_name` is implemented twice, with different trimming in `NodeStatus.file_server_name`.

Here, `.rstrip` call is done, but in not in `NodeStatus.file_server_name`. This means the rule and the trimming policy is different depending which `file_server_name` you call. I would recommend to only keep one implementation if possible. Also, I couldn't find where `file_server_name` is used.

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,101 @@
...
+
+      # Default number of attempts before giving up.
+      DEFAULT_RETRIES = 3
+
+      # One entry in the returned name table.
+      #
+      # @!attribute [r] name [String] the NetBIOS name (trimmed)
+      # @!attribute [r] suffix [Integer] 1-byte NetBIOS suffix
+      # @!attribute [r] group [Boolean] true for a group name, false for unique
+      # @!attribute [r] active [Boolean] true if the name is registered
```
cdelafuente-r7, line 26: I might be wrong, but I don't think the `Struct` attributes are read-only ('[r]')?

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,101 @@
...
+      # @param host [String] target IP address
+      # @param kwargs [Hash] forwarded to {.query}
+      # @return [String, nil]
+      def self.file_server_name(host, **kwargs)
+        entries = query(host, **kwargs) or return nil
+        entry = entries.find { |e| e.suffix == 0x20 && e.unique? }
+        entry&.name
+      end
+
+      # @!visibility private
```
cdelafuente-r7, line 88: `@!visibility private` only affects YARD output. If this method really needs to be private, I think using `private_class_method :entries_from` after the definition might work.

spec/lib/ruby_smb/nbss/node_status_response_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      expect(group_entry.name_flags.group.to_i).to eq(1)
+      expect(group_entry.name_flags.active.to_i).to eq(1)
+      expect(group_entry.group?).to be true
+      expect(unique_entry.name_flags.group.to_i).to eq(0)
+      expect(unique_entry.unique?).to be true
+    end
+  end
+
+  describe '#file_server_name' do
+    it 'returns the name with suffix 0x20 and the unique bit clear' do
```
cdelafuente-r7, line 78: I believe it is the group bit that is cleared here. Not sure what the `unique bit` is though?

```suggestion
    it 'returns the name with suffix 0x20 and the group bit clear' do
```

lib/ruby_smb/nbss/node_status_request.rb
```diff
@@ -0,0 +1,31 @@
...
+      uint16                    :qdcount,        label: 'QDCount', initial_value: 1
+      uint16                    :ancount,        label: 'ANCount', initial_value: 0
+      uint16                    :nscount,        label: 'NSCount', initial_value: 0
+      uint16                    :arcount,        label: 'ARCount', initial_value: 0
+
+      # Question section. For a node status query this is always the wildcard
+      # NetBIOS name (16 bytes of 0x2A / 0x00), L1-encoded.
+      netbios_name :question_name,  label: 'Question Name'
```
cdelafuente-r7, line 26: According to this comment, It might be a good idea to use the wildcard NetBIOS name as a default value for this field.

```suggestion
      netbios_name :question_name, label: 'Question Name', initial_value: '*'.ljust(16, "\x00").freeze
```
You can also use a constant at the beginning of the class for self documentation, if you prefer:
```ruby
  # RFC 1002 4.2.17: a node status query always asks for the wildcard name.
  WILDCARD_NAME = '*'.ljust(16, "\x00").freeze
```
```suggestion
      netbios_name :question_name, label: 'Question Name', initial_value: WILDCARD_NAME
```

Then, you delete line 51 in `node_status.rb` (`request.question_name.set('*'.ljust(16, "\x00"))`) and the `before` hook line 6-8 in `node_status_request_spec.rb` specs.

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,101 @@
...
+
+          data, addr = udp_socket.recvfrom(4096)
+          next if data.nil? || data.empty?
+          # Reject replies from an unexpected source — this is a unicast
+          # query, so anything not from `host` is spoofed or stray.
+          next unless addr.nil? || addr[3].nil? || addr[3] == host
+
+          response = NodeStatusResponse.read(data)
+          # Reject replies whose transaction ID doesn't match our request.
+          next unless response.transaction_id.to_i == request.transaction_id.to_i
```
cdelafuente-r7, line 66: It's a first good step to make sure the response is the one we want, but I would add more check to make sure it is really the good one (untested):
- checking if it is a response packet
- checking if `rr_type` if a `QUESTION_TYPE_NBSTAT`

```suggestion
          next unless response.transaction_id.to_i == request.transaction_id.to_i
          next unless response.flags.response.to_i == 1
          next unless response.rr_type.to_i == NodeStatusRequest::QUESTION_TYPE_NBSTAT
```

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        retries.times do
+          udp_socket.send(bytes, 0, host, port)
+          next unless IO.select([udp_socket], nil, nil, timeout)
+
+          data, addr = udp_socket.recvfrom(4096)
+          next if data.nil? || data.empty?
+          # Reject replies from an unexpected source — this is a unicast
+          # query, so anything not from `host` is spoofed or stray.
+          next unless addr.nil? || addr[3].nil? || addr[3] == host
```
cdelafuente-r7, line 62: If `host` is a hostname, `send` will try to resolve it (line 55), but since `addr[3]` from `recvfrom` is always the numeric peer address, the comparison here will always fail (`addr[3] == host`). We cannot handle the responsibility of resolving any potential hostname here, especially if it is a `Rex::Socket::Udp` and we are pivoting. This must be the caller responsibility.

So, we need to make sure `host` is in the same numerical form. I suggest to document this and add some checks to verify `host` is in the correct IPv4 form (e.g. `10.0.0.1`). Note that NBNS doesn't support IPv6, so we don't need to handle it.

For example (untested):
```ruby
def self.query(host, port: NBNS_PORT, timeout: DEFAULT_TIMEOUT,
         retries: DEFAULT_RETRIES, udp_socket:)
  expected_address = begin
    IPAddr.new(host).native
  rescue IPAddr::InvalidAddressError
    raise ArgumentError, "host must be an IPv4 address, got #{host.inspect}"
  end
  unless expected_address.ipv4?
    raise ArgumentError, "NBNS is IPv4-only, got #{host.inspect}"
  end
  ...
```
Then we have to make sure the comparison is correct:

```suggestion
          source = addr.is_a?(Array) ? addr[3] : nil
          unless source.nil?
            source_address = begin
              IPAddr.new(source).native
            rescue IPAddr::InvalidAddressError
              nil
            end
            next unless source_address == expected_address
          end
```

I would also update the method documentation to make sure the caller understand what kind of `udp_socket` is expected:
- Must implement `UDPSocket#send(mesg, flags, host, port)` and `#recvfrom(maxlen)`
- Leave `do_not_reverse_lookup` at its default so `recvfrom` does not perform a reverse DNS lookup per datagram

cdelafuente-r7 (commented):
Thanks @Z6543. I reviewed this PR and added some comments and suggestions. Please, let me know if you need any help.

lib/ruby_smb/nbss/name_service_header_flags.rb
Z6543: Done. Split the header's second word into the RFC 1002 §4.2.1.1 groups: `NameServiceHeaderFlags` now holds only the 7 NM_FLAGS bits, with new `NameServiceOpcode` (R + OPCODE) and `NameServiceResultCode` (RCODE) records. Verified BinData packs the three nested records contiguously into the 2-byte word (e.g. `0x8415`).

lib/ruby_smb/nbss/node_status_request.rb
Z6543: Done, as suggested — the request/response headers now declare `opcode` / `nm_flags` / `rcode`.

lib/ruby_smb/nbss/node_status.rb
Z6543: Good catch. Moved the `rescue` inside `retries.times`, so a malformed datagram on one attempt no longer aborts the remaining retries. Added a spec: "retries after a malformed datagram and succeeds on a later attempt".

lib/ruby_smb/nbss/node_status.rb
Z6543: Done — `transaction_id` now uses `SecureRandom.random_number(0x10000)` (full 0–0xFFFF range), added `require 'securerandom'`, and updated the specs to stub `SecureRandom`.

lib/ruby_smb/nbss/negative_session_response.rb
Z6543: Done — moved the error-code constants to `lib/ruby_smb/nbss.rb` next to the session packet types. `error_msg` still resolves them via lexical scope; updated the `client_spec` refs to `RubySMB::Nbss::<CONST>`.

lib/ruby_smb/client.rb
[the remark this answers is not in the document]
Z6543: You're right, and the PR description was stale. The design changed in commit "Let the caller own NetBIOS name-resolution reconnect": `session_request` no longer retries internally. On `0x82` it raises `NetBiosSessionService` carrying the numeric `error_code`; the caller resolves the real name via `NodeStatus.file_server_name`, reconnects (the server drops the socket after a negative response), and retries — so the new socket follows the same route (e.g. a Metasploit pivot). This is documented at `client.rb:669–677`. Added a spec asserting `CALLED_NAME_NOT_PRESENT` propagates with the right `error_code`, and corrected the PR description.

lib/ruby_smb/nbss/node_status_response.rb
Z6543: Removed `NodeStatusResponse#file_server_name`; kept the module-level `NodeStatus.file_server_name` (the advertised entry point referenced in `client.rb`). Neither was called elsewhere.

lib/ruby_smb/nbss/node_status.rb
Z6543: Fixed — Struct members are read-write; changed the YARD attributes to `[rw]`.

lib/ruby_smb/nbss/node_status.rb
Z6543: Done — replaced with `private_class_method :entries_from` (also applied to the new `source_matches?` helper). Internal `self.` calls still work.

spec/lib/ruby_smb/nbss/node_status_response_spec.rb
Z6543: That `describe` block was removed along with `NodeStatusResponse#file_server_name` (see the dedup comment), so the wording no longer exists. Equivalent coverage lives in the `NodeStatus.file_server_name` specs.

lib/ruby_smb/nbss/node_status_request.rb
Z6543: Done — added `WILDCARD_NAME` constant, set it as `question_name`'s `initial_value`, and dropped both the `request.question_name.set(...)` line and the spec `before` hook.

lib/ruby_smb/nbss/node_status.rb
Z6543: Done — the loop now also rejects non-responses (`opcode.response != 1`) and replies whose `rr_type` isn't `QUESTION_TYPE_NBSTAT`, using `next` so a stray packet still allows retries.

lib/ruby_smb/nbss/node_status.rb
Z6543: Done — `query` now validates `host` is a numeric IPv4 address (raises `ArgumentError` otherwise) and compares the normalized `IPAddr` of the source against the target (extracted into `source_matches?`). Documented the `udp_socket` contract (`send`/`recvfrom` signatures, leave `do_not_reverse_lookup` at default). Added an `ArgumentError` spec.

lib/ruby_smb/nbss/node_status_request.rb
```diff
@@ -0,0 +1,38 @@
...
+      # 16 bytes of 0x2A ('*') padded with 0x00.
+      WILDCARD_NAME = '*'.ljust(16, "\x00").freeze
+
+      endian :big
+
+      # 12-byte NBNS header (RFC 1002 4.2.1.1 and 4.2.1.2).
+      uint16                    :transaction_id, label: 'Transaction ID'
+      name_service_opcode       :opcode,         label: 'Opcode'
+      name_service_header_flags :nm_flags,       label: 'Flags'
+      name_service_result_code  :rcode,          label: 'Result Code'
```
cdelafuente-r7: After thinking about this again, I found that my suggested changes were not the best way to do it. `nm_flags` are good, but I don't think is worth having specific classes for `opcode` and `result_code`, since it is just one or two field and the call to get the field values would not be very intuitive:
```ruby
request = NodeStatusRequest.new
puts request.opcode.opcode
puts request.opcode.response
puts request.rcode.rcode
```
I would suggest to only keep the `name_service_header_flags` class and include the other fields directly, sorry about this.

```suggestion
      bit1                      :response,       label: 'Response', initial_value: 0
      bit4                      :opcode,         label: 'Opcode',   initial_value: 0
      name_service_header_flags :nm_flags,       label: 'Flags'
      bit4                      :rcode,          label: 'Result Code', initial_value: 0
```
This would give you this:
```ruby
request = NodeStatusRequest.new
puts request.opcode
puts request.response
puts request.rcode
```

lib/ruby_smb/nbss/node_status_response.rb
```diff
@@ -0,0 +1,86 @@
...
+    # Does not decode the trailing STATISTICS field; callers only need the
+    # name table.
+    class NodeStatusResponse < BinData::Record
+      endian :big
+
+      # 12-byte NBNS header.
+      uint16                    :transaction_id, label: 'Transaction ID'
+      name_service_opcode       :opcode,         label: 'Opcode'
+      name_service_header_flags :nm_flags,       label: 'Flags'
+      name_service_result_code  :rcode,          label: 'Result Code'
```
cdelafuente-r7: Same suggestion I left in `NodeStatusRequest` class.

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,136 @@
...
+            response = NodeStatusResponse.read(data)
+            # Reject anything that isn't the response to our own query.
+            next unless response.transaction_id.to_i == request.transaction_id.to_i
+            next unless response.opcode.response.to_i == 1
+            next unless response.rr_type.to_i == NodeStatusRequest::QUESTION_TYPE_NBSTAT
+
+            return entries_from(response)
+          rescue IOError, EOFError, SystemCallError
+            next
+          end
```
cdelafuente-r7: Since the `rescue` is for the whole `retries` block, a new embedded `begin` block is not necessary.

```suggestion
          udp_socket.send(bytes, 0, host, port)
          next unless IO.select([udp_socket], nil, nil, timeout)

          data, addr = udp_socket.recvfrom(4096)
          next if data.nil? || data.empty?
          next unless source_matches?(addr, expected_address)

          response = NodeStatusResponse.read(data)
          # Reject anything that isn't the response to our own query.
          next unless response.transaction_id.to_i == request.transaction_id.to_i
          next unless response.opcode.response.to_i == 1
          next unless response.rr_type.to_i == NodeStatusRequest::QUESTION_TYPE_NBSTAT

          return entries_from(response)
        rescue IOError, EOFError, SystemCallError
          next
```

lib/ruby_smb/nbss/node_status.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+      end
+      private_class_method :entries_from
+
+      # Accept a reply only when it comes from the queried address. `recvfrom`
+      # returns the numeric peer address, which is compared against the
+      # normalized target. A nil/unknown source (e.g. a mock socket) is allowed.
+      def self.source_matches?(addr, expected_address)
+        source = addr.is_a?(Array) ? addr[3] : nil
+        return true if source.nil?
```
cdelafuente-r7: I understand this exists to accommodate specs doubles that return `[data, nil]` (e.g. `expect(udp_sock).to receive(:recvfrom).and_return(["\xff\xff".b, nil])`), but a real `recvfrom` always returns the 4-element `Addrinfo` array.

I would prefer not having this and treat a nil source as non-matching. Having a nil value means either it is not an Array or `addr[3]` is nil; in both cases, something went wrong and it cannot be trusted. The main goal of all of this is to make sure the source is there and is the expected value.
```suggestion
        return false if source.nil?
```
The method documentation should also be updated, explaining that a UDP socket #recvfrom always returns the sender as `[address_family, port, host, numeric_host]`, so the numeric address lives at `addr[3]`. Any datagram whose source cannot be determined or does not match the queried target is rejected.

Finally the specs will need to return a stub `Addrinfo` array instead of `[data, nil]` (e.g `expect(udp_sock).to receive(:recvfrom).with(4096).and_return([response_bytes, ['AF_INET', 137, '10.0.0.2', '10.0.0.2']])`).

lib/ruby_smb/client.rb
```diff
@@ -668,6 +668,14 @@ def wipe_state!
 
     # Requests a NetBIOS Session Service using the provided name.
     #
+    # On refusal the raised {RubySMB::Error::NetBiosSessionService} carries the
+    # numeric NBSS `error_code`. A `0x82` (CALLED_NAME_NOT_PRESENT) rejection of
```
cdelafuente-r7: Consider adding a YARD link for this in the docs instead of hardcoding the value:
```suggestion
    # numeric NBSS `error_code`. A {RubySMB::Nbss::CALLED_NAME_NOT_PRESENT} rejection of
```

cdelafuente-r7 (commented):
Thanks @Z6543 for these great additions! I've tested using the SMB share password enumeration module you also wrote (CVE-2000-0979) and it works great. I left a few comments for you to review.

lib/ruby_smb/nbss/node_status_request.rb
Z6543: Done — reverted the split. Both headers now declare `response` / `opcode` / `nm_flags` / `rcode` inline (keeping only `NameServiceHeaderFlags`), and I deleted `NameServiceOpcode` / `NameServiceResultCode`. Access is now `request.opcode` / `request.response` / `request.rcode`.

lib/ruby_smb/nbss/node_status_response.rb
Z6543: Done — same inline layout applied to `NodeStatusResponse`.

lib/ruby_smb/nbss/node_status.rb
Z6543: Done — removed the embedded `begin`; the `rescue IOError, EOFError, SystemCallError` now attaches directly to the `retries.times` block.

lib/ruby_smb/nbss/node_status.rb
Z6543: Done — `source_matches?` now `return false if source.nil?`, so a datagram whose source cannot be determined is treated as non-matching. Updated the method doc (a UDP `#recvfrom` returns the sender as `[address_family, port, host, numeric_host]`, so the numeric address is `addr[3]`; any datagram whose source cannot be determined or does not match the queried target is rejected) and switched the specs to return a stub Addrinfo array (`['AF_INET', 137, '10.0.0.2', '10.0.0.2']`) instead of `[data, nil]`.

lib/ruby_smb/client.rb
Z6543: Done — replaced the hardcoded `0x82` with the `{RubySMB::Nbss::CALLED_NAME_NOT_PRESENT}` YARD link.

cdelafuente-r7 (approved):

PR #297, opened 2026-04-24
Add #set_unix_link to the SMB1 tree

zeroSteiner:
This closes a gap with the Rex SMB client. There will be an accompanying PR to Metasploit that leverages it by updating the `auxiliary/smb/samba_symlink_traversal` module.


jheysel-r7 (approved):
Thanks for the PR @zeroSteiner. Testing on the framework side was as expected.

TIL about `Transaction`, `Transaction2` and their subcommands. Also that they only exist in SMB1 and not SMB2 or 3.

 This appears to be implement what's needed on the framework side 👍

